PR #4, opened 2017-09-01
Add contributing guidelines and external licenses subdirectory

namurphy:
I copied over the [contributing guidelines from PlasmaPy](https://github.com/PlasmaPy/PlasmaPy/blob/master/CONTRIBUTING.md) and adapted them to fiasco.  It's ready for review.

I also created a `licenses` directory at the top-level and put the PlasmaPy license in it, since that's where I copied the file from.  This directory would also be a good place to put things like licenses from ChiantiPy, Astropy, SunPy, and so on for whenever code from a different open source repository gets adapted into fiasco.


wtbarnes:
The contributing guidelines look great. Thanks for adding that.

The one thing I don't like about putting the license files in a separate directory is that GitHub does not recognize that your repo is licensed and doesn't give you the little badge on the top left of the banner on the repo. I'd rather keep the single license in the main root of the repo until we add additional licenses and need a directory to keep them in.

wtbarnes:
@namurphy I've just started using the sunpy package template which uses a `licenses` directory and the BSD-3 license so that part is not needed. Could you remove the licenses part of this PR and then I'll merge the contributing guidelines?

namurphy:
@wtbarnes - This should be all set, but let me know if there's anything else I need to do.

wtbarnes:
Thanks @namurphy. FYI I've also started using the atomic module in plasmapy for conversions between element name, atomic symbol, and atomic number. It has made things a lot easier.

namurphy:
That's great to hear!  One thing I was wondering about was how fiasco would handle atomic notation.  I'm not so much a fan of notation like `"fe_16"` or `"Fe XVI"` since the off-by-one difference from the charge state always confuses me.  I personally much prefer something like "`Fe 16+`" since that has the charge state built right in.  At the same time, most astronomers are used to `"Fe XVI"`.  Anyway, separate question, but just thought I'd mention it.

wtbarnes:
I think actually I will start a separate issue about this...

PR #53, opened 2020-02-21
Updates to be compatible with astropy 4

wtbarnes:
This PR makes several updates, mostly to be compatible with astropy v4 and the way quantities are passed through numpy functions. 

**TODO: the ionization equilibrium calculation, at least for Fe, is very wrong! Need to understand exactly why this is...**


wtbarnes:
RE above comment, the ioneq problem was due to an issue in the direct ionization cross section calculation due to the way I had stripped units off of a quantity

PR #58, opened 2020-02-27
Add unit to wavelength return

### LINKED ISSUE #57: Add units to wavelengths returned by .spectrum()
Currently `IonCollection.spectrum()` returns a tuple. The first element is wavelengths, but they don't currently have units and are just a bare array. Units should be added.

dstansby:
Closes #57. I also added a warning where the width of the broadening kernel is assumed to be 0.1 Angstrom, so it's obvious to the user what's happening.

### DIFF
--- a/fiasco/collections.py
+++ b/fiasco/collections.py
@@ -1,6 +1,8 @@
 """
 Multi-ion container
 """
+import warnings
+
 import numpy as np
 import astropy.units as u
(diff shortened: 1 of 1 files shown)


wtbarnes:
Thanks for catching this! As you can tell, that function is a bit of a mess. I essentially slapped it together for my PyAstro talk in 2018 and haven't really touched it since. It definitely needs some love.

PR #59, opened 2020-02-27
Add returns block to level_populations() doc

dstansby:
It would be nice eventually if this 3D array came with indexes (ie. an xarray? a 3D astropy table?)

### DIFF
--- a/fiasco/ion.py
+++ b/fiasco/ion.py
@@ -258,9 +258,9 @@ Using Datasets:
         Returns
         -------
         `~astropy.units.Quantity`
-            A ``(l, m, n)`` shaped quantity, where ``l`` is the number number of
+            A ``(l, m, n)`` shaped quantity, where ``l`` is the number of
             temperatures, ``m`` is the number of densities, and ``n``
-            is the energy level.
+            is the number of energy levels.
         """
         level = self._elvlc['level']
         lower_level = self._scups['lower_level']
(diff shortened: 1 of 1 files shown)


fiasco/ion.py
```diff
@@ -254,6 +254,13 @@ def level_populations(self,
         density : `~astropy.units.Quantity`
         include_protons : `bool`, optional
             If True (default), include proton excitation and de-excitation rates
+
+        Returns
+        -------
+        `~astropy.units.Quantity`
+            A ``(l, m, n)`` shaped quantity, where ``l`` is the number number of
+            temperatures, ``m`` is the number of densities, and ``n``
+            is the energy level.
```
wtbarnes:
```suggestion
            A ``(l, m, n)`` shaped quantity, where ``l`` is the number of
            temperatures, ``m`` is the number of densities, and ``n``
            is the number of energy levels.
```

wtbarnes:
Thanks for this! I fixed a minor typo. I didn't actually mean to commit it (:sweat_smile:), but it was a very small change.

dstansby:
Oh I think I committed it, presumably straight after you left it!

wtbarnes:
> It would be nice eventually if this 3D array came with indexes (ie. an xarray? a 3D astropy table?)

Good point. Returning a 3D array without any context as to what the dimensions correspond to is not very user friendly. An xarray is a good idea. That could be useful for the contribution function, level population, intensity too...

PR #60, opened 2020-02-27
Allow scalar density to level_populations

### LINKED ISSUE #56: Allow scalar density input to ion.level_populations?
Currently doing something like `ion.level_populations(5e8 * u.cm**-3)` errors with
```python
----> 1 ions[0].level_populations(5e8 * u.cm**-3)

~/github/fiasco/fiasco/util/decorators.py in func_wrapper(*args, **kwargs)
     20                 return default
     21             else:
---> 22                 return func(*args, **kwargs)
     23         return func_wrapper
     24     return decorator

~/miniconda3/lib/python3.8/site-packages/astropy/units/decorators.py in wrapper(*func_args, **func_kwargs)
    232             # Call the original function with any equivalencies in force.
    233             with add_enabled_equivalencies(self.equivalencies):
--> 234                 return_ = wrapped_function(*func_args, **func_kwargs)
    235             if wrapped_signature.return_annotation not in (inspect.Signature.empty, None):
    236                 return return_.to(wrapped_signature.return_annotation)

~/github/fiasco/fiasco/ion.py in level_populations(self, density, include_protons)
    290         # Populate density dependent terms and solve matrix equation for each density value
    291         populations = np.zeros(self.temperature.shape + density.shape + (level.max(),))
--> 292         for i, d in enumerate(density):
    293             c_matrix = coeff_matrix.copy()
    294             # Collisional excitation and de-excitation out of current state

~/miniconda3/lib/python3.8/site-packages/astropy/units/quantity.py in __iter__(self)
   1017     def __iter__(self):
   1018         if self.isscalar:
-> 1019             raise TypeError(
   1020                 "'{cls}' object with a scalar value is not iterable"
   1021                 .format(cls=self.__class__.__name__))

TypeError: 'Quantity' object with a scalar value is not iterable
```

I think this is just a case of making sure a scalar input is broadcast to at least 1D so that the for loop works.

dstansby:
Fixes #56

### DIFF
--- a/fiasco/conftest.py
+++ b/fiasco/conftest.py
@@ -1,6 +1,7 @@
 # this contains imports plugins that configure py.test for astropy tests.
 # by importing them here in conftest.py they are discoverable by py.test
 # no matter how it is invoked within the source tree.
+import os
 import pytest
 from astropy.tests.plugins import *
 
@@ -13,6 +14,8 @@ TEST_FILES = [
     'chianti.ip',
     'chianti.ioneq',
     'h_1.elvlc',
+    'h_1.wgfa',
+    'h_1.scups',
     'h_2.rrparams',
     'he_1.elvlc',
     'he_2.elvlc',
--- a/fiasco/ion.py
+++ b/fiasco/ion.py
@@ -297,6 +297,7 @@ Using Datasets:
                 upper_level_p, level, dex_rate_p.value.T, 0).T * dex_rate_p.unit
 
         # Populate density dependent terms and solve matrix equation for each density value
+        density = np.atleast_1d(density)
         populations = np.zeros(self.temperature.shape + density.shape + (level.max(),))
         for i, d in enumerate(density):
             c_matrix = coeff_matrix.copy()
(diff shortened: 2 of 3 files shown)


wtbarnes:
Thanks! Would you be willing to add a test here too? Just something as simple as doing the level populations given a scalar density and checking that the shape of the resulting matrix is as expected. I've found that 0 II is a good test ion as it is quick to compute. The Fe ones are all quite slow.

No worries if you don't have the time. I'm happy to merge this as is.

dstansby:
Yep, will add a test

wtbarnes:
Might want to wait until #65 is merged and do a rebase before adding a test. I refactored the tests a bit.

wtbarnes:
I made your density test check the resulting shape of the level populations explicitly and separated out the test for the scalar temperature.

wtbarnes:
Ah sorry, Because my commit was the last one and I squash-merged, it looks like you don't show up in the commit history at all for this PR 🤦‍♂ 

I'm not sure the best way to go about preserving authorship when squash merging, but I'll be more careful about this in the future.

Issue #26, opened 2018-04-21
Speedup level populations calculation


wtbarnes:
The level populations calculation is extremely computationally expensive and scales poorly with number of available transitions. This means that level populations for ions with very large number of transitions and available energy levels (e.g. Fe IX, XI) take a **very** long time to compute.

The level populations calculation is essentially needed for everything so poor performance here impacts many other calculations. Maybe something like numba or cython could be used to speedup this calculation or it could be parallelized using something like dask.

wtbarnes:
Replace linalg.solve with SVD as in the ionization equilibrium calculation. They are really the same equation.

An equivalent function is implemented in [Dask](http://docs.dask.org/en/latest/array-api.html#dask.array.linalg.svd) so this could be potentially parallelized as well.

kdere:
speeding things up would be very helpful.
right now I am trying to get Version 9 of the database and ChiantiPy out.  There are significant changes in the way autoionization states are handled

wtbarnes:
After some profiling, it seems the main bottleneck here is computing the effective collision strengths from the scups data, specifically the descaling all of the collision strengths.

dstansby:
Just to keep a record, here's the profile I've just run on `level_populations`:
```python
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   244                                               @needs_dataset('elvlc', 'wgfa', 'scups')
   245                                               @u.quantity_input
   246                                               @profile
   247                                               def level_populations(self,
   248                                                                     density: u.cm**(-3),
   249                                                                     include_protons=True) -> u.dimensionless_unscaled:
   250                                                   """
   251                                                   Compute energy level populations as a function of temperature and density
   252                                           
   253                                                   Parameters
   254                                                   ----------
   255                                                   density : `~astropy.units.Quantity`
   256                                                   include_protons : `bool`, optional
   257                                                       If True (default), include proton excitation and de-excitation rates
   258                                                   """
   259         1       2326.0   2326.0      0.0          level = self._elvlc['level']
   260         1       4734.0   4734.0      0.0          lower_level = self._scups['lower_level']
   261         1       3802.0   3802.0      0.0          upper_level = self._scups['upper_level']
   262         1        407.0    407.0      0.0          coeff_matrix = np.zeros(self.temperature.shape + (level.max(), level.max(),))/u.s
   263                                           
   264                                                   # Radiative decay out of current level
[...]

PR #69, opened 2020-03-02
Cache list of available ions

dstansby:
Goes a long way to fixing #26. This reduces a `level_populations` calculation from 36 seconds to 25 seconds for me

### DIFF
--- a/fiasco/fiasco.py
+++ b/fiasco/fiasco.py
@@ -1,6 +1,7 @@
 """
 Package-level functions
 """
+import functools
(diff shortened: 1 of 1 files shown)


wtbarnes:
:tada: 

This is not surprising to me at all though I think this will only give you a speedup when including the proton excitation/deexcitation rates. The `proton_to_electron_ratio` calculation that is called when calculating the proton density needed for the proton rates requires looping through every ion in the database

PR #75, opened 2020-03-05
Misc speedups

dstansby:
- Lazily evaluate names in `Ion` so they aren't calculated every time an `Ion` is created.
- Move a multiplication out of a for loop

### DIFF
--- a/fiasco/base.py
+++ b/fiasco/base.py
@@ -19,17 +19,11 @@ class Base(object):
     """
 
     def __init__(self, ion_name, hdf5_dbase_root=None, **kwargs):
-        element, ion = ion_name.split()
+        self._element, ion = ion_name.split()
         if '+' in ion:
             ion = f"{int(ion.strip('+')) + 1}"
-        self.atomic_number = plasmapy.atomic.atomic_number(element.capitalize())
-        self.element_name = plasmapy.atomic.element_name(element.capitalize())
-        self.atomic_symbol = plasmapy.atomic.atomic_symbol(element.capitalize())
         self.ionization_stage = int(ion)
         self.charge_state = self.ionization_stage - 1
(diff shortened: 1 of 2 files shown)


fiasco/fiasco.py
```diff
@@ -86,7 +86,8 @@ def proton_electron_ratio(temperature: u.K, **kwargs):
             if ioneq is None:
                 warnings.warn(f'Not including {ion.ion_name}. Ionization fraction not available.')
                 continue
-            denominator += ioneq * abundance * ion.charge_state
+            denominator += ioneq * ion.charge_state
+        denominator *= abundance
```
wtbarnes: I don't think this is equivalent. This multiplies the abundance for that element on all existing terms. E.g. if you had summed over H and then did the sum over He, the He abundance would get multiplied by the H terms which is not correct.

dstansby: Ah yes, will revert this commit.

wtbarnes:
I like the idea of moving things into properties and I certainly see how this would speed up ion creation which would be great for collections. I'm wondering though if this will slow things down in other places, i.e. if I'm using `self._ion_name` in a function repeatedly, I'm now repeatedly calling `split`, `plasmapy.atomic`, etc.

Let me investigate a bit...

dstansby:
I did originally cache them too, but that didn't seem to go too well last time! In python 3.8 there's `cached_propery` that might help. (https://docs.python.org/3/library/functools.html#functools.cached_property)

wtbarnes:
So it doesn't look like these are really used in too many places, at least not enough to make a difference performance wise. The one exception might be the atomic number, but it is easy enough to just evaluate that once (i.e. `Z=self.atomic_number`) as needed.

PR #76, opened 2020-03-05
Do a manual caching of list_ions

### LINKED ISSUE #73: Mangled function signature when using lru_cache
#69 wrapped `fiasco.list_ions` with the `functools.lru_cache` decorator to greatly increase its performance. This has the unfortunate side affect of messing up the autogenerated API docs. For some reason, automod-api cannot sort out the current module or correction function name.

Is there something we can do about this? maybe wrapping the decorator? [This blog post](https://hynek.me/articles/decorators/) has some interesting related information. Should this be reported upstream to the sphinx-automodapi extension?

dstansby:
Fixes #73. Also fixed an astropy sphinx deprecation warning.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -72,7 +72,7 @@ except ImportError:
             sys.path.insert(1, a_h_path)
 
 # Load all of the global Astropy configuration
-from astropy_helpers.sphinx.conf import *
+from sphinx_astropy.conf import *
 
 # Get configuration information from setup.cfg
 try:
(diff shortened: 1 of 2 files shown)


wtbarnes:
Hmm the tests really do not seem to like this...

Also, I'm a little uneasy about having a global variable that stores the ion_list. What if someone creates an ion using a different database? They'll then be checking against the ion list for the wrong file.

I really appreciate you working on this. I know it is kind of a stupid bottleneck.

I'm wondering now if an easy way to do this would be just to store a list of all the ions in the database as a separate entry in the database at the top level. That way, there's no need to crawl the whole thing every time you want list every ion. Could do the same with the element list too.

Issue #74, opened 2020-03-04
Caching the test database


dstansby:
As far as I can tell, running the tests attemptes to download and build a smaller version of the database in a temporary folder. If it could detect an already-present installation and use that instead it would save having to do this every time local tests are run.

wtbarnes:
I intentionally made this change recently. I wanted to make the setup for the database on the CI easier and to avoid the test results being dependent on the state of how the database is configured on a given machine.

Building the minimal HDF5 version of the database takes no time at all, but downloading the tarball of the ASCII files every time can be a bit annoying (and probably is not very respectful to the CHIANTI webpage). In my mind, this should still be the default behavior, but I wonder if there is a way to cache this with pytest? 

Another option would just be to check the rc file (as is done when instantiating an ion) and download it there, though it feels strange to put a test-dependent configuration option in the rc file.

dstansby:
Caching the download seems like the sensible option, with a SHA to check it's the right file.

wtbarnes:
I wonder if the new remote data manager in sunpy would work for this…

wtbarnes:
@dstansby are you happy with #79 and #86 as a solution to this?

dstansby:
Yep 👍

PR #79, opened 2020-03-06
Detect a local copy of the database when testing

dstansby:
See #74. This detects if a local copy of the ASCII database is present, and if so uses that to build the h5 database needed for testing.

Perhaps there should be SHA checks on the files to check they are as expected?

### DIFF
--- a/fiasco/conftest.py
+++ b/fiasco/conftest.py
@@ -6,7 +6,7 @@ import pytest
 from astropy.tests.plugins import *
 
 from fiasco.util import build_hdf5_dbase, download_dbase
-from fiasco.util.setup_db import CHIANTI_URL, LATEST_VERSION
+from fiasco.util.setup_db import CHIANTI_URL, LATEST_VERSION, FIASCO_HOME
 
 # Minimal set of CHIANTI files needed to run the tests
 TEST_FILES = [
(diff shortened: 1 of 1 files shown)


fiasco/conftest.py
```diff
@@ -33,6 +33,12 @@
 
 @pytest.fixture(scope='session')
 def ascii_dbase_root(tmpdir_factory):
+    # If we already have a local copy, just return the path to that
+    local_dbase_path = os.path.join(FIASCO_HOME, 'chianti_dbase')
```
wtbarnes: `fiasco.defaults` stores the settings in the `fiascorc` file so we should probably check there instead of hardcoding here. It will default to `FIASCO_HOME / 'chianit_dbase'` anyway.

wtbarnes:
I'm not entirely opposed to this approach. It just makes me a little uneasy that it makes the tests dependent on the state of the database on a given machine. Either way, I do agree that we shouldn't be downloading every time the tests are run locally 😆

Maybe we could just add an entry to the `fiascorc` file (`test_ascii_dbase_path`) and use that to check for a local copy?

Is there an entrypoint for passing commandline arguments to pytest? or some kind of rc file? That would be preferable to mixing test setup with the more user-focused settings in the `fiascorc` file.

dstansby:
I was thinking to run an SHA check on all the ascii files being included in the test database - that way they're guaranteed to be the same each time.

fiasco/conftest.py
wtbarnes: I'll just merge this and fix this on #86

PR #81, opened 2020-03-07
Speedup listing of all ions in database

wtbarnes:
Addresses #69 #73 ping @dstansby 

This adds an `ion_index` entry at the top of the database. `fiasco.list_ions` will then just pull from this rather than searching the entire database tree to create the list.

Some comparisons when creating an `Ion`: 

* original (no index, no cache)
  ```python
  %timeit fiasco.base.IonBase('Fe 18')
  36.3 ms ± 327 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
  ```
* with #69 (i.e. LRU cache)
  ```python
  %timeit fiasco.base.IonBase('Fe 18',)
  113 µs ± 601 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
  ```
* this PR
  ```python
  %timeit fiasco.base.IonBase('Fe 18',)
  1.25 ms ± 4.93 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
  ```

This is ~30x faster than the original method, but still about ~10x slower than the cache method. However, I can still create a collection of every ion in the database in <1 s using this PR so I'm not sure that further optimization is that necessary?

### DIFF
--- a/fiasco/fiasco.py
+++ b/fiasco/fiasco.py
@@ -32,21 +32,23 @@ def list_elements(hdf5_dbase_root, sort=True):
     return elements
 
 
-@functools.lru_cache()
+#@functools.lru_cache()
 def list_ions(hdf5_dbase_root, sort=True):
     """
     List all available ions in the CHIANTI database
     """
-    ions = []
-    root = DataIndexer.create_indexer(hdf5_dbase_root, '/')
-    for f in root.fields:
-        try:
-            el = plasmapy.atomic.atomic_symbol(f.capitalize())
-            for i in root[f].fields:
-                if f == i.split('_')[0]:
-                    ions.append(f"{el} {i.split('_')[1]}")
-        except InvalidParticleError:
-            continue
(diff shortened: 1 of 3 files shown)


dstansby:
The overhead in creating Ions really kicks in when every element and then subsequently every ion is created (~5000 I think, so 5000 calls to `Ion.__init___`). This PR seems like a sensible way to go 👍

wtbarnes:
Out of curiosity, in what context are you creating 5000 ions? There are only about 495 unique ions in the database which is the example I mentioned in the above comment. If there are use cases for creating many more than this, the overhead may be more than my original estimate.

dstansby:
Ah I think I'm an order of magnitude out - 500 sounds more like it... With your timings up top before this change it took 18 seconds to create all 500, which was a significant slow down to `level_populations`.

wtbarnes:
Ok, I'll go ahead and merge this then. If you find it is still a bottleneck for some use cases, we could always go back to the lru_cache method.

PR #86, opened 2020-03-16
Verify hashes of files used for testing

dstansby:
This should check that the files being used/downloaded for testing are exactly as expected.

### DIFF
--- a/docs/conftest.py
+++ b/docs/conftest.py
@@ -1,24 +1,33 @@
 # NOTE: this file is duplicated here because doctest cannot pick up custom
 # fixtures when the conftest file is anywhere else for some reason
+import os
+
 import pytest
 
+import fiasco
 from fiasco.util import build_hdf5_dbase, download_dbase
 from fiasco.util.setup_db import CHIANTI_URL, LATEST_VERSION
 
 # Minimal set of CHIANTI files needed to run the doctests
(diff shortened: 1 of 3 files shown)


dstansby:
I think this has worked as expected! I have Chianti 8.0.6 locally, which I generated the hashes with, but the tests expect 8.0.7, so this failed. Will update the hashes to 8.0.7 in a new commit.

dstansby:
I think this has worked as expected! I have Chianti 8.0.6 locally, which I generated the hashes with, but the tests expect 8.0.7, so this failed. Will update the hashes to 8.0.7 in a new commit.

wtbarnes:
This is great! I should merge #79 first right?

dstansby:
I don't think it matters which order they go in.

dstansby:
To save myself the hassle of downloading the database again and again I've rebased this on top of #79, so that should probalby now be merged first.

wtbarnes:
I rebased out the commits from #79 and made changed the check for the ascii database fixture to check a separate variable in the rc file, `test_ascii_dbase_root` (the naming of all of these variables should probably be revisited 😅 ). That way, your "production" database and the one that you're testing against don't collide. Otherwise, if you were using a version other than 8.0.7, the tests would then all fail because the hashes are for that specific version.

PR #90, opened 2020-04-18
Add transitions test

### LINKED ISSUE #88: Add convenience property for ion.transitions.wavelength[~ion.transitions.is_twophoton]
`ion.transitions.wavelength[~ion.transitions.is_twophoton]` is the way to get the correspond wavelengths for the contribution function, but this is a bit cumbursome. It would be nice if `Ion` had a property that returned this. Not sure what it should be called though...

### LINKED ISSUE #101: Add len() to Transitions
It would be helpful if the `Transitions` class defined `__len__`, to get the number of transitions stored.

dstansby:
Fixes #101

### DIFF
--- a/fiasco/level.py
+++ b/fiasco/level.py
@@ -55,6 +55,9 @@ class Transitions(object):
         self._elvlc = elvlc
         self._wgfa = wgfa
 
+    def __len__(self):
+        return len(self._wgfa['wavelength'])
+
     @property
     def is_twophoton(self):
         """
--- a/fiasco/tests/test_ion.py
+++ b/fiasco/tests/test_ion.py
@@ -160,10 +160,19 @@ def test_radd_ions(ion, another_ion):
     assert collection[1] == ion
     assert collection[0] == another_ion
 
+
 def test_transitions(ion):
     trans = ion.transitions
     assert isinstance(trans, fiasco.Transitions)
-    assert len(trans) == 1
+    assert len(trans) == 361
+    # These values have not been tested for correctness
+    assert not trans.is_twophoton[0]
+    assert trans.is_observed[0]
+    assert u.allclose(trans.A[0], 0.000155 / u.s)
+    assert u.allclose(trans.wavelength[0], 703729.75 * u.Angstrom)
+    assert u.allclose(trans.upper_level[0], 2)
+    assert u.allclose(trans.lower_level[0], 1)
+    assert u.allclose(trans.delta_energy[0], 2.82273956e-14 * u.erg)
 
 
 def test_create_ion_without_units_raises_units_error(hdf5_dbase_root):
--- a/fiasco/tests/test_ion.py
+++ b/fiasco/tests/test_ion.py
@@ -160,6 +160,11 @@ def test_radd_ions(ion, another_ion):
     assert collection[1] == ion
     assert collection[0] == another_ion
 
+def test_transitions(ion):
+    trans = ion.transitions
+    assert isinstance(trans, fiasco.Transitions)
+    assert len(trans) == 1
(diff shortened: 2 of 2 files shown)


wtbarnes:
I'm not opposed to the name. A bit awkward maybe, but it is explicit.

Also, if you rebase this, the tests should pass.

wtbarnes:
One minor comment: for any instances where the not two-photon transitions are selected, could you replace it with this new property? I think the only place is in the contribution function, but I maybe forgetting others.

fiasco/ion.py
```diff
@@ -92,6 +92,12 @@ def __radd__(self, value):
     def transitions(self):
         return Transitions(self._elvlc, self._wgfa)
 
+    @property
+    @needs_dataset('elvlc', 'wgfa')
+    def non_twophoton_transitions(self):
+        all_transitions = self.transitions
+        return all_transitions[~all_transitions.is_twophoton]
```
wtbarnes, line 99: Somehow I just now noticed this but how exactly does this work? There is no __getitem__ implemented on Transitions so what is it slicing here exactly?

dstansby: Haha oh dear I'm not sure. Let me give this another go (and add tests!!!)

wtbarnes: Not opposed to having getitem on Transitions btw. That could be interesting.

fiasco/level.py
```diff
@@ -55,6 +55,12 @@ def __init__(self, elvlc, wgfa):
         self._elvlc = elvlc
         self._wgfa = wgfa
 
+    def __getitem__(self, idx):
+        elvc = self._elvlc[idx]
```
wtbarnes, line 59: I don't think this is the right approach. The `_elvlc` and `_wgfa` properties are `DataIndexer` objects

https://github.com/wtbarnes/fiasco/blob/07f14fe1a5b4273bb2bb0743b612c6d31cd00632/fiasco/io/datalayer.py#L35

and their `__getitem__` is meant for accessing columns on those datasets, e.g. the 'E_obs', the observed energies on `_elvlc`.

wtbarnes:
I think you rebased out your `non_twophoton_transitions` property 😅

dstansby:
In the spirit of just getting something working, this is now just a test, but for some reason the transitions on the test ions is just returning `None`... Any ideas?

dstansby:
Not sure what happened here, but I've rebased down to the simple transitions test which should be fine to merge.

fiasco/tests/test_ion.py
```diff
@@ -160,6 +160,11 @@ def test_radd_ions(ion, another_ion):
     assert collection[1] == ion
     assert collection[0] == another_ion
 
+def test_transitions(ion):
+    trans = ion.transitions
+    assert isinstance(trans, fiasco.Transitions)
+    assert len(trans) == 1
```
wtbarnes, line 166: Why is the expected result 1?

dstansby: 🤷 not sure, I think this is just a placeholder for now. The test isn't passing currently, and is requesting user input which pytest doesn't like. I'm guessing it's because there is a file missing somewhere.

wtbarnes: I'm confused. All the failures I see on the CI are because Fe 5 has 361 transitions and not 1.

dstansby: 🤣 I got confused with https://github.com/wtbarnes/fiasco/pull/104

dstansby: Okay, should be fixed now, with added bonus tests!

PR #92, opened 2020-04-21
Default to normal database for test database

dstansby:
If `test_ascii_dbase_root` isn't set in the config file, default to the normal database. This way the files automatically get picked up, but the test database can still be configured to another database if desired by setting it in the fiascorc file.


wtbarnes:
I'm not sure I want this to be the default behavior. By default, I'd like to test on the minimal version of the database directly downloaded from the CHIANTI webpage, and then optionally cached by the user. This way, the test suite is, by default, testing against a known version of the database (because we pin the URL and version).

This also effectively short circuits the logic `conftest.py` that downloads a new copy if the `test_ascii_dbase_path` is not set as now this value is always set.

https://github.com/wtbarnes/fiasco/blob/7e9e3b9eba992955a91cb46a113d01aa80e38e3c/fiasco/conftest.py#L46-L55

dstansby:
Fair, sounds like I should just spend 10 seconds setting my config 😆

PR #95, opened 2020-04-28
Add some value checks to test_ion.py

dstansby:
I've added some checks that contain actual values, so if a computation gets changed it's caught in the test suite. None of these values are checked for correctness.

### DIFF
--- a/fiasco/tests/test_ion.py
+++ b/fiasco/tests/test_ion.py
@@ -47,12 +47,16 @@ def test_scalar_temperature(hdf5_dbase_root):
     ion = fiasco.Ion('H 1', 1 * u.MK, hdf5_dbase_root=hdf5_dbase_root)
     ioneq = ion.ioneq
     assert ioneq.shape == (1,)
+    assert np.isnan(ioneq)
 
 
 def test_scalar_density(hdf5_dbase_root):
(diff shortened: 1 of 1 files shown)


fiasco/tests/test_ion.py
```diff
@@ -47,12 +47,16 @@ def test_scalar_temperature(hdf5_dbase_root):
     ion = fiasco.Ion('H 1', 1 * u.MK, hdf5_dbase_root=hdf5_dbase_root)
     ioneq = ion.ioneq
     assert ioneq.shape == (1,)
+    assert np.isnan(ioneq)
```
wtbarnes, line 50: You writing this test has already caught a bug!!!! This should not be a NaN! PR coming soon! I'll rebase and fix this value.

wtbarnes: In this case, we can actually test approximately for correctness as `ioneq` is just interpreting from the supplied data.

wtbarnes:
That ioneq nan test should now fail (it fails locally for me). Let me add an actual consistency test with the ioneq data that we interpolate from and you can see what you think.

Everything else looks good to me!

dstansby:
Looks 👍  to me

wtbarnes:
This PR is the perfect example of why you need tests! 😄

PR #99, opened 2020-04-28
Add proton collision tests

dstansby:
I had to add a new ion here, because it needed a `.psplups` file. Perhaps this could replace some another ion?


wtbarnes:
I think adding a new one is fine. Most of the "weight" of the database is in a few of the Fe `.scups` files so as long as we don't start pulling those I think we are ok. 

At some point down the road, we may want to come back and clean up the test files a bit.

PR #102, opened 2020-04-29
Add more value tests

dstansby:
I've gone with continuing to include required files for `Fe 5` here.


dstansby:
🎉 it went from red to orange!

PR #103, opened 2020-04-29
Test emmisivity

dstansby:
I thought I would open failing tests on top of #102 in individual PRs (that can hopefully have fixes applied to them). This one is for `emmisivity`.

### DIFF
--- a/fiasco/ion.py
+++ b/fiasco/ion.py
@@ -425,6 +425,7 @@ Using Datasets:
         --------
         contribution_function : Calculate contribution function, :math:`G(n,T)`
         """
+        density = np.atleast_1d(density)
         g = self.contribution_function(density, **kwargs)
         return g * (density**2)[np.newaxis, :, np.newaxis]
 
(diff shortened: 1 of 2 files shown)


fiasco/ion.py
```diff
@@ -416,7 +416,7 @@ def emissivity(self, density: u.cm**(-3), **kwargs) -> u.erg * u.cm**(-3) / u.s:
         contribution_function : Calculate contribution function, :math:`G(n,T)`
         """
         g = self.contribution_function(density, **kwargs)
-        return g * (density**2)[np.newaxis, :, np.newaxis]
```
dstansby, line 419: I'm not sure what these `newaxis` bits were doing here?

wtbarnes: Hmm I think because `g` has dimensions (`ntemperture`, `ndensity`, `ntransitions`)? I'm confused as to how this works now that you've removed those `newaxis` calls. Maybe I just don't understand numpy broadcasting at all? My understanding was that in order to do broadcast like this, the trailing dimension of each array had to have the same length.

dstansby: Yep, you're right, just needed a `atleast_1d` in case density is a scalar.

PR #104, opened 2020-04-29
Free bound test

dstansby:
Similar to #103, this fails.


wtbarnes:
I think this was failing because in `Ion.free_bound`, another ion instance is created for the recombining ion. When that ion was created, it was not using the test database and was trying to download and setup a new copy (hence the asking for input). It also requires the `fblvl` file for that recombining ion. 

I've fixed it such that the recombining ion instance that is created always uses the same version of the database as the ion instance that creates it (this was actually a bigger bug and I'm glad this caught it!). I've also added the `fe_6.fblvl` file to the test database.

Issue #118, opened 2020-11-05
Improve error when .h5 database not present


dstansby:
```python
No HDF5 database found at /Volumes/Work/chianti.h5. Build it now? [Y/n] 
No CHIANTI database found at /Volumes/Work/chianti. Download it from http://www.chiantidatabase.org/download/CHIANTI_8.0.7_data.tar.gz? [y/N] 
Traceback (most recent call last):
  File "examples/response_function.py", line 14, in <module>
    ion = Ion('H 1+', Te)
  File "/Users/dstansby/github/astropy/astropy/units/decorators.py", line 251, in wrapper
    return_ = wrapped_function(*func_args, **func_kwargs)
  File "/Users/dstansby/github/fiasco/fiasco/ion.py", line 48, in __init__
    super().__init__(ion_name, *args, **kwargs)
  File "/Users/dstansby/github/fiasco/fiasco/base.py", line 32, in __init__
    if self.ion_name not in fiasco.list_ions(self.hdf5_dbase_root, sort=False):
  File "/Users/dstansby/github/fiasco/fiasco/fiasco.py", line 40, in list_ions
    ions = root['ion_index']
  File "/Users/dstansby/github/fiasco/fiasco/io/datalayer.py", line 100, in __getitem__
    if key not in self:
  File "/Users/dstansby/github/fiasco/fiasco/io/datalayer.py", line 93, in __contains__
    with h5py.File(self.hdf5_dbase_root, 'r') as hf:
  File "/Users/dstansby/miniconda3/envs/dev/lib/python3.8/site-packages/h5py/_hl/files.py", line 424, in __init__
    fid = make_fid(name, mode, userblock_size,
  File "/Users/dstansby/miniconda3/envs/dev/lib/python3.8/site-packages/h5py/_hl/files.py", line 190, in make_fid
    fid = h5f.open(name, flags, fapl=fapl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5f.pyx", line 96, in h5py.h5f.open
OSError: Unable to open file (unable to open file: name = '/Volumes/Work/chianti.h5', errno = 2, error message = 'No such file or directory', flags = 0, o_flags = 0)
```

This should give a much nicer error about the file not being present.

PR #119, opened 2020-11-05
Add more checks for database existence

dstansby:
Fixes https://github.com/wtbarnes/fiasco/issues/118. I just dumped a check before every read call. Maybe there's a nicer way to do this?


wtbarnes:
Hmm is this really more informative than a `FileNotFoundError` that would crop up when trying to read the HDF5 file?

wtbarnes:
Could the check be done once in the `DataIndexerHDF5` constructor?

dstansby:
I made a property with the check, which is a bit more code, but a bit more general (in case someone deletes or moves their database during a session (although I have no idea why they would do that ...))

PR #122, opened 2020-11-05
Add a sphinx gallery

### LINKED ISSUE #109: Example gallery?
Should the documentation include an example gallery? Currently, I've relied on a more monolithic approach when it comes to the tutorial, but maybe it could be supplemented with an example gallery showing how to do common analysis tasks with CHIANTI.

dstansby:
Fixes #109 

No content yet, but should still build


wtbarnes:
Nice! I'm happy to merge this as is, or if you wanted to add a few more examples or refactor anything. Some of the things in the tutorial could probably go in here.

Overhauling the docs is an issue outside the scope of this PR so I wouldn't want to hold it up.

dstansby:
I think this is 👍 from my point of view, but I just want to add some more comments to the example.

dstansby:
Should be good to go if the tests pass

PR #124, opened 2020-11-05
Don't default to None in required datasets decorator

### LINKED ISSUE #89: Derived quantities should error more gracefully when abundance is None
Received this error when trying to calculate the contribution function for Cr 9 for the 211 channel. Original used 'sun_coronal_1992_feldman' as the abundance file, also tried 'grevesse_photospheric_1998' - neither worked. 

```
Traceback (most recent call last):

  File "<ipython-input-475-5f00bd24d0c3>", line 1, in <module>
    K_nT, K_T, KT_dict, ch, response, temperature = temp_resp(211, 'chromium 9')

  File "/Users/shirman/Desktop/DEM.py", line 46, in temp_resp
    g = ion_name.contribution_function(density)

  File "/Users/shirman/fiasco/fiasco/util/decorators.py", line 22, in func_wrapper
    return func(*args, **kwargs)

  File "/Users/shirman/anaconda3/lib/python3.7/site-packages/astropy/units/decorators.py", line 234, in wrapper
    return_ = wrapped_function(*func_args, **func_kwargs)

  File "/Users/shirman/fiasco/fiasco/ion.py", line 384, in contribution_function
    term = np.outer(self.ioneq, 1./density) * self.abundance * 0.83

  File "/Users/shirman/anaconda3/lib/python3.7/site-packages/astropy/units/quantity.py", line 947, in __mul__
    return super().__mul__(other)

  File "/Users/shirman/anaconda3/lib/python3.7/site-packages/astropy/units/quantity.py", line 481, in __array_ufunc__
    result = super().__array_ufunc__(function, method, *arrays, **kwargs)

TypeError: unsupported operand type(s) for *: 'float' and 'NoneType'
```

### LINKED ISSUE #93: Don't silently return None when a value can't be computed
Currently the `needs_dataset` silently (eek!) returns `None` if a dataset isn't needed. This confused me for a while, and I think it should at least warn. Looking through the places where this decorator is used, I think it might be reasonable to error too if a dataset isn't present.

e.g. if I don't have the files needed to calculate `ion.transitions`, calling `ion.transitions` should probably raise an error. I think #89 is a subset of this issue for specific case of abundances.

dstansby:
Fixes #93 , Fixes #89 

Returning `None` here is wild, and just leads to very confusing error messages down the line. It's especially wild in a decorator. Given the `default` argument isn't used anywhere, remove this, and error if any of the datasets aren't present.

Could perhaps have a nicer error message with the ion name in it?

### DIFF
--- a/fiasco/collections.py
+++ b/fiasco/collections.py
@@ -10,6 +10,7 @@ from astropy.modeling.models import Gaussian1D
 import plasmapy
 
 import fiasco
+from fiasco.util.exceptions import MissingDatasetException
 
 __all__ = ['IonCollection']
 
@@ -92,9 +93,12 @@ class IonCollection(object):
         free_bound = u.Quantity(np.zeros(self.temperature.shape + wavelength.shape),
                                 'erg cm^3 s^-1 Angstrom^-1')
         for ion in self:
-            fb = ion.free_bound(wavelength, **kwargs)
-            # Not valid for all ions
-            if fb is not None:
+            try:
+                fb = ion.free_bound(wavelength, **kwargs)
+            except MissingDatasetException:
+                # TODO: log the skipped ion
+                continue
+            else:
                 free_bound += fb * ion.abundance * ion.ioneq[:, np.newaxis]
         return free_bound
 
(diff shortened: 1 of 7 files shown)


wtbarnes:
I do think this is the best option. I've made some suggestions in #93 that I think would work well here. There are a few places in the code (in the methods on `Ion`) that would need to be fixed to account for this not returning `None` but it should be easy to fix.

wtbarnes:
Ok so I perhaps unwisely went down the rabbit hole of the `None` behavior. Issuing the `MissingDatasetException` in the decorator is definitely the right call. I've gone a step further and force a `KeyError` at the data interface layer if an entry is missing. Though a bit messier, this is better than a bunch of `None`s flying around all over the place.

PR #127, opened 2020-11-09
Add level populations test with protons

dstansby:
Should add a bit more coverage. I had to add a new ion to the test database as it needed a `psplups` file, but I removed the `fe_10.psplups` one as I think that was only there for testing the parser.

### DIFF
--- a/fiasco/conftest.py
+++ b/fiasco/conftest.py
@@ -1,0 +1,80 @@
+# this contains imports plugins that configure py.test for astropy tests.
+# by importing them here in conftest.py they are discoverable by py.test
+# no matter how it is invoked within the source tree.
+import os
+import pytest
+from astropy.tests.plugins import *
+
+import fiasco
+from fiasco.util import build_hdf5_dbase, download_dbase
+from fiasco.util.setup_db import CHIANTI_URL, LATEST_VERSION, FIASCO_HOME
+
+# Minimal set of CHIANTI files needed to run the tests
+TEST_FILES = {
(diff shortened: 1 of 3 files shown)


fiasco/conftest.py
```diff
@@ -48,8 +48,15 @@
     'fe_18.rrparams':                       '6a36e5db1e377d0c30652960eb77b93d',
     'fe_18.drparams':                       '1e8b0194aeb099b2653508110e388200',
     'fe_27.rrparams':                       '75383b0f1b167f862cfd26bbadd2a029',
-    'fe_10.psplups':                        'dd34363f6daa81dbf106fbeb211b457d',
```
wtbarnes, line 51: This file is used in the `test_proton_collision` test.

wtbarnes: If you want, just replace the `fe10` ion fixture with `mg5` and then use that for the proton collision rate and the new level population test. There's no reason to have both.

fiasco/ion.py
```diff
@@ -290,7 +292,8 @@ def level_populations(self,
         try:
             _ = self._psplups
         except KeyError:
-            # TODO: log this
+            warnings.warn('psplups data not available, not including protons '
```
wtbarnes, line 295: I'm now wondering whether this is the right approach, or if this should just be removed and an error thrown in the `if include_protons` block by just putting the `proton_collision_excitation_rate` call first. This of course means if you're iterating over a number of ions, you're constantly having to catch this exception so maybe this is best how it is.

dstansby: I'll get rid of this for now, I think it snuck in as I wanted to double check what files I needed to add.

PR #128, opened 2020-11-10
Test multiple versions of the database

wtbarnes:
Fixes #128 

This PR is a first pass at running the tests on multiple versions of the CHIANTI database. This can stay open for a while and should be rebased as compatibility fixes are made for past and future versions of the database.


wtbarnes:
After a first pass, it looks like the v7 tests blow up because of the use of `_scups` files. This isn't too surprising.

It is surprising to me that the v9 tests pass. However, I guess that we aren't introducing any of the new files here and the code has not been changed in any way to accommodate the new features in v9. This does confirm that the code should "just work" with newer versions of the database, even if it doesn't actually utilize the v9 data yet.

wtbarnes:
Closing this as the CI set up is now quite different and there is now a v9 test in place.

PR #142, opened 2021-02-25
Update chianti d/l url

dstansby:


wtbarnes:
Wow. I don't know how I missed that this had happened.

PR #144, opened 2021-02-25
Cache database download

dstansby:


wtbarnes:
I'm all for cacheing and speeding up the tests, but I'm just a little lost. Could you explain what is going on in the GitHub actions script?

.github/workflows/run-tests.yml
```diff
@@ -42,13 +47,22 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+    - name: Cache chianti database
+      uses: actions/cache@v2
+      if: matrix.toxenv != 'build_docs' && matrix.toxenv != 'codestyle'
```
dstansby, line 52: This line says only run the caching on the test runs, and not the doc or codestyle runs.

.github/workflows/run-tests.yml
```diff
@@ -42,13 +47,22 @@ jobs:
...
+      env:
+        # Bump the version number when changing the chianti version to test with
+        cache-name: cache-chianti-dbase-807
```
dstansby, line 55: This an environment variable set to the name of the cache, it's only used in line 58 below

.github/workflows/run-tests.yml
```diff
@@ -42,13 +47,22 @@ jobs:
...
+      with:
+        path: ~/.chianti
```
dstansby, line 57: This is the directory that's cached.

.github/workflows/run-tests.yml
```diff
@@ -42,13 +47,22 @@ jobs:
...
+        key: ${{ env.cache-name }}-${{ matrix.toxenv }}
```
dstansby, line 58: And this is the key that it's cached based on, ie. if this key changes then the old cache is deleted and replaced by a new version. I've just set it to a custom name, and the name of the tox environment, as the chianti database only changes when the chianti version changes.

.github/workflows/run-tests.yml
```diff
@@ -42,13 +47,22 @@ jobs:
...
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: python -m pip install --upgrade tox codecov
     - name: Run tox
-      run: python -m tox -e  ${{ matrix.toxenv }}
+      run: python -m tox -e  ${{ matrix.toxenv }} ${{ matrix.cache_cmd }}
```
dstansby, line 66: This has just added the cache_cmd environment variable to be passed to the tox command, which sets where the database is downloaded to.

dstansby:
I've tried to annotate the changelog, but let me know if I missed a bit you don't understand, or any of my explanations don't make sense.

wtbarnes:
> I've tried to annotate the changelog, but let me know if I missed a bit you don't understand, or any of my explanations don't make sense.

Thanks! I think I get it now. I think I was just thrown off by the GH cache action. Thanks for doing this. I think this will be much kinder to the CHIANTI webpage 😅

Issue #120, opened 2020-11-05
numpy deprecation warning


dstansby:
Should probably fix this:

```python
/Users/dstansby/github/fiasco/fiasco/util/tools.py:175: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray
  nots = np.array(list(map(splrep_szero, x, y)))
```

dstansby:
FYI I'm still getting this warning:
```python
/Users/dstansby/github/fiasco/fiasco/util/tools.py:155: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
  x = np.asarray(x)
```

wtbarnes:
Ah I think I understand. The warning is popping up in a different place than last time.

The reason this isn't popping up in the tests is because we are not testing the Fe XII ion. The array in question is only ragged (and thus only triggers this warning) for some ions. The fits over the scaled collision strengths may or may not all be the same length and this varies per ion.

The solution to suppressing this is to add some more sanitization to the inputs to `burgess_tully_descale`.

PR #146, opened 2021-02-25
Error on warnings when testing

dstansby:
Fixes https://github.com/wtbarnes/fiasco/issues/120

### DIFF
--- a/fiasco/ion.py
+++ b/fiasco/ion.py
@@ -203,7 +203,7 @@ Using Datasets:
 
         .. math::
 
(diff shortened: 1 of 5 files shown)


dstansby:
@wtbarnes I think this is good for the tests - I can fix docs in another PR

wtbarnes:
This looks great to me. Thanks! If you wanted to follow up with a docs fix PR that would be great. From the failures, it looks like just a dependency issue.

PR #165, opened 2021-11-22
Use `cached_property`

### LINKED ISSUE #164: Cached properties on `Element` object
The `ionization_equilibrium` and `rate_matrix` properties on the `Element` object should use `cached_property` instead of `property`. They can be somewhat expensive to compute and it does not make sense to recompute them each time. This feature is only available in 3.8 and higher, but I don't see that being an issue.

https://docs.python.org/3.8/library/functools.html#functools.cached_property

wtbarnes:
Fixes #164 and also bumps minimum Python version to 3.8

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -16,8 +16,8 @@ jobs:
         include:
         # Test the oldest and newest configuration on Mac and Windows
         - os: macos-latest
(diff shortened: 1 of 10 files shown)


wtbarnes:
This should probably be done wherever possible on a lot of the `Ion` prooperties and methods.

PR #167, opened 2022-05-09
Fix free-free continuum and add IDL comparison tests

wtbarnes:
This PR adds a set of tests that can perform systematic comparisons with the equivalent CHIANTI IDL functions and also fixes the free-free continuum emissivity.

## TODO

- [x] Fix differences between IDL and fiasco free-free emissivity
- [x] Rebase
- [x] Finish docstring with equation to free-free method
- [x] Fix codestyle

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,7 +32,7 @@ repos:
       - id: check-ast
       - id: check-case-conflict
       - id: trailing-whitespace
-        exclude: ".*(.fits|.fts|.fit|.txt)$"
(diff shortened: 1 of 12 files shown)


wtbarnes:
There was an issue when comparing with the free-free continuum emission as computed by IDL. The issue was that fiasco was using the atomic number instead of the charge to compute the free-free emission. After this fix, the comparison is now,

image

There is now a test that asserts that the difference between the two is <.5%, a tolerable amount given that the two are computed using fairly different approaches.

PR #168, opened 2022-05-29
Add example for computing non-equilibrium ion populations

### LINKED ISSUE #163: Add functions for calculating ionization fraction as a function of time
It would be convenient to have two functions that return the population fractions as a function of time for a given time, temperature, and (optionally) density profile under the assumptions of equilibrium and non-equilibrium ionization.

For the equilibrium calculation, it is just a matter of interpolating the temperatures to the associated equilibrium values:

```python
ioneq = el.equilibrium_ionization()
ioneq_interp = interp1d(el.temperature.value, ioneq.value.T,  kind='cubic',fill_value='extrapolate')(T.value).T
```

The non-equilbrium case is slightly more complicated, but still reasonably easy to implement:

https://github.com/wtbarnes/synthesizAR/blob/b6dfc3a71f7a63d46458df534709f50694246dac/synthesizAR/atomic/chianti.py#L21-L70

wtbarnes:
Fixes #163 

This adds an example gallery entry for computing the time-dependent ion population fractions.


wtbarnes:
Fixes #163

PR #175, opened 2022-11-07
Add `codespell` to pre-commit

namurphy:
This pull request adds [codespell](https://github.com/codespell-project/codespell) to the pre-commit configuration.  

Instead of searching for words that don't show up in a dictionary, codespell looks for words that match a dictionary of common misspellings.  Consequently, codespell leads to many fewer false positives than more traditional spellcheckers applied to code, but it will not catch all misspellings.  In this repo, the only false positive was `te`.

I added a codespell configuration section to `setup.cfg`, including which files and words to ignore.

I originally attempted to add the configuration to `pyproject.toml`, but when running `pre-commit run --all-files`, it was not picking up the configuration there, even though it did so when running codespell regularly.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,3 +39,8 @@ repos:
         exclude: ".*(.fits|.fts|.fit|.txt|tca.*)$"
       - id: mixed-line-ending
         exclude: ".*(.fits|.fts|.fit|.txt|tca.*)$"
+
+  - repo: https://github.com/codespell-project/codespell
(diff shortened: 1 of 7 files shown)


wtbarnes:
Great! Thanks for this. I often find spelling mistakes in examples, docstrings after the fact and then forget to fix them so this is great!

namurphy:
Not a problem!  I recently made some codespell PRs for Astropy as a means of procrastination-driven development (PDD) and was happy to add it here too.

PR #176, opened 2022-11-07
Include Python 3.11 in test suite

namurphy:
This PR makes two changes to `tox.ini` in preparation for a v0.1.0 release:
 - Add a test for Python 3.11 to the test suite
 - ~~Change the Python 3.9 test to `windows-latest`~~

Because PlasmaPy depends on Numba (at least for now), and fiasco depends on PlasmaPy, it'll be necessary to wait until Numba works on Python 3.11.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -36,6 +36,10 @@ jobs:
           python-version: '3.10'
           cache_cmd: -- --dbase-download-dir ~/.chianti
           toxenv: py310
+        - os: ubuntu-latest
+          python-version: '3.11'
+          cache_cmd: -- --dbase-download-dir ~/.chianti
+          toxenv: py311
         # Docs
         - os: ubuntu-latest
           python-version: 3.9
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -42,11 +42,11 @@ jobs:
           toxenv: py311
         # Docs
         - os: ubuntu-latest
-          python-version: 3.9
+          python-version: '3.11'
           toxenv: build_docs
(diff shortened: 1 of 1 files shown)


namurphy:
> Windows test failure is because when reading the fiasco rc file, [it is expected that the $HOME environment variable exists](https://github.com/wtbarnes/fiasco/blob/acee6b957e4bcbebe871902c293facc973a5210e/fiasco/util/util.py#L12) which is probably bad practice in general. Perhaps we could take a more general approach to reading this configuration file?

Agreed!  I'll raise an issue on this.

> The 3.11 failure is because numba is being picked up as a dependency but numba is not yet compatible with 3.11. However, I'm not sure why numba is getting picked up as a dependency at all. It is not an explicit dependency of fiasco.

That's PlasmaPy's fault here, since we have Numba as a requirement.  I've been dreaming of dropping it because of how long it takes for Numba to get released after a few Python version.  Hopefully before next year!

wtbarnes:
Ah gotcha. I didn't realize it had been picked up as a dependency in plasmapy. Is there room to make it an optional dependency?

namurphy:
> Ah gotcha. I didn't realize it had been picked up as a dependency in plasmapy. Is there room to make it an optional dependency?

We use it within some of our formulary functions, so we'd have to do some rewriting to make it optional.  My own preference would be to cythonize the numba out of everything so that we drop it as a dependency entirely.

namurphy:
Just downscoped this to remove the tests for Windows, which are going to be covered in a separate PR (see also #179).

And still waiting for Numba to work on Python 3.11...

wtbarnes:
This is superseded by #245

PR #180, opened 2022-11-20
Update pre-commit configuration and apply changes

namurphy:
This PR makes a few changes to `.pre-commit-config.yaml`, `setup.cfg`, and `pyproject.toml` for the `pre-commit` configuration and linter/autoformatter tools used from PlasmaPy.  While this PR touches a lot of files, only the ones mentioned above (and `.gitignore`) were manually changed.  All other changes were automagically made by `pre-commit` 

 - I updated `.pre-commit-config.yaml` to most recent version of `flake8` in pre-commit from the GitHub repo (as opposed to the old GitLab link, which was weirdly asking me for a login)
 - I added a config for `pyupgrade` to pre-commit for Python 3.8+, excluding fancy `__future__` annotation stuff.  `pyupgrade` is a neat tool that updates existing code to use newer/recommended language features
 - I added a pre-commit hook for `codespell`, and set it to ignore `te` which is used as electron temperature.
 - I added a pre-commit hook for `absolufy-imports`, which changes relative imports to absolute imports so that they're more readable, and it's easier to move files to different locations if needed (i.e. into a subdirectory).
 - I added a pre-commit hook for `isort`, which sorts import statements. I included the settings to make `isort` consistent with `black`, just in case `black` gets added to `pre-commit` later on.


pyproject.toml
```diff
@@ -46,3 +46,16 @@ build-backend = 'setuptools.build_meta'
...
+wrap_length = 80
+sections = ["FUTURE", "STDLIB", "FIRSTPARTY", "LOCALFOLDER"]
+known_first_party = ["fiasco", ]
+default_section = "STDLIB"
+multi_line_output = 3
+use_parentheses = true
+include_trailing_comma = true
+force_alphabetical_sort_within_sections = true
+honor_noqa = true
+lines_between_types = 1
```
namurphy, line 61: This is the configuration needed to make isort compatible with black.

wtbarnes:
this looks great to me! The doc build fail is because of an aiapy issue so I won't worry about that. Could you apply the isort fixes here so that the codestyle check passes?

PR #182, opened 2022-11-21
Use `pathlib` instead of `os.path` and enable working with Windows

### LINKED ISSUE #179: Fix issue with getting fiasco to run on Windows
In #176, I attempted to add Windows to the GitHub Action for running tests.  It resulted in the following error:

```python
ImportError while loading conftest 'D:\a\fiasco\fiasco\conftest.py'.
..\..\conftest.py:6: in <module>
    from fiasco.util import build_hdf5_dbase, download_dbase
..\..\fiasco\__init__.py:4: in <module>
    from .util.util import setup_paths
..\..\fiasco\util\__init__.py:4: in <module>
    from .util import *
..\..\fiasco\util\util.py:12: in <module>
    FIASCO_HOME = os.path.join(os.environ['HOME'], '.fiasco')
C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\os.py:679: in __getitem__
    raise KeyError(key) from None
E   KeyError: 'HOME'
```

One possible approach to take while fixing this would be to [replace usage of `os.path` with `pathlib`](https://builtin.com/software-engineering-perspectives/python-pathlib).  In `pathlib`, paths are represented as `Path` instances rather than raw strings, which makes it more pythonic.

namurphy:
This initial goals of this PR are:
 - Fix the bug with running fiasco on Windows (#179)
 - Migrate `os.path` → `pathlib`

Closes #179.  I previously tried Windows tests on #176.


namurphy:
To get home directory via `pathlib` in an OS independent way:
```Python
>>> import pathlib
>>> pathlib.Path.home()
PosixPath('/home/namurphy')
```
Yay!

docs/conf.py
```diff
@@ -178,11 +176,11 @@
...
         ]
     )
-    with open(os.path.join(FIASCO_HOME, 'fiascorc'), 'w') as f:
+    with open(fiasco_home / 'fiascorc', 'w') as f:
         c = configparser.ConfigParser()
         c.add_section('database')
-        c.set('database', 'ascii_dbase_root', ascii_dbase_root)
-        c.set('database', 'hdf5_dbase_root', hdf5_dbase_root)
+        c.set('database', 'ascii_dbase_root', str(ascii_dbase_root))
+        c.set('database', 'hdf5_dbase_root', str(hdf5_dbase_root))
```
namurphy, line 183: This isn't really Pythonic, but using `str` here is because `ConfigParser.set()` doesn't work with `Path` instances.

namurphy:
Finally got it!  The issue was that I was doing `Path.mkdir()`, but without `parents=True` for `~/.fiasco/tmp`.  It was working locally because `.fiasco` existed for me previously, but failing in GitHub Actions because it creating an entirely new environment.

fiasco/util/util.py
```diff
@@ -72,18 +74,22 @@ def get_masterlist(ascii_dbase_root):
...
+    # TODO: Replace usage with pathlib, noting that pathlib does not
+    # have a direct equivalent to os.walk
+
     skip_dirs = ['version_3', 'deprecated', 'masterlist', 'ioneq', 'dem', 'ancillary_data', 'ip',
                  'abundance', 'continuum', 'instrument_responses']
     # List of all files associated with ions
     ion_files = []
     for root, sub, files in os.walk(ascii_dbase_root):
-        if not any([sd in root for sd in skip_dirs]) and not any([sd in sub for sd in skip_dirs]):
+        if all(sd not in root for sd in skip_dirs) and all(sd not in sub for sd in skip_dirs):
```
namurphy, line 86: This was a suggested change from Sourcery.

namurphy:
Given my very roundabout commit history, doing a "squash and merge" seems like the best option to me.

wtbarnes:
Thanks for making this massive effort! This looks great to me. The CI issue was an especially tricky one.

PR #190, opened 2022-11-25
Enable tests with Windows and begin `os.path` → `pathlib` transition

namurphy:
This is my second attempt to fix an problem with running tests on Windows.  I'm attempting this in a different PR because I kept getting a temporary path not found error in #182 that I had trouble tracking down.  I had tests passing on my workstation, but an error on GitHub.  I'm going to try to track down the exact changes that I made that broke this.

This supersedes #182. Probably.

Closes https://github.com/wtbarnes/fiasco/issues/179. I previously tried Windows tests on https://github.com/wtbarnes/fiasco/pull/176.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -14,14 +14,23 @@ jobs:
       fail-fast: false
       matrix:
         include:
-        # Test the oldest and newest configuration on Mac and Windows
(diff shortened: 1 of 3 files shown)


namurphy:
Nevermind!  Going back to #182.  

The issue was that when doing `path.mkdir()`, I had to set `parents=True` to make the parent directory exist also.

PR #196, opened 2022-12-05
Fix free-bound continuum function

### LINKED ISSUE #42: Free-bound continuum does not match IDL output
See #32

wtbarnes:
Fixes #42 

This PR is an effort to fix the inconsistency between the free-bound continuum implementation in fiasco versus is what is implemented in CHIANTI IDL. This also adds a suite of tests for comparing against the IDL implementation.

TODOs:

- [x] Update free-bound regression tests
- [x] Add tests for comparing free-bound per ion against IDL results
- [x] Finish docstring for free-bound method on Ion
- [x] Update docstring for free-bound method on Collection

Current progress:

image
image

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,7 +37,7 @@ repos:
       - id: debug-statements
       - id: check-added-large-files
       - id: end-of-file-fixer
-        exclude: ".*(.fits|.fts|.fit|.txt|.bib|tca.*)$"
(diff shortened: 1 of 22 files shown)


wtbarnes:
I found two bugs that have FINALLY resolved these inconsistencies:

- The free-bound continuum was being incorrectly multiplied by the population fraction of the recombined ion. It is now corectly multiplied by the recombined ion.
- The free-bound level energies were not being correctly filled in by the theoretical energies when the observed energies were missing

The updated comparison now looks like the following:

image
image

PR #200, opened 2022-12-12
Use OA template for CI

### LINKED ISSUE #178: Add a GitHub Action to automate the release process
There are a few GitHub Actions out there that automate the release process to PyPI.  For example, @StanczakDominik has been working on https://github.com/PlasmaPy/PlasmaPy/pull/1779, which I'm really excited about.  We could add a GitHub Action like that one to fiasco to facilitate quicker releases.

wtbarnes:
Fixes #178 

Simplify CI by using the OA template

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -6,72 +6,24 @@ on:
   pull_request:
     branches: [ main ]
 
+
 jobs:
-  build:
-
-    runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-        # Test the oldest and newest configuration on Mac
-        - os: macos-latest
-          python-version: 3.8
-          toxenv: py38
-          cache_cmd: -- --dbase-download-dir ~/.chianti
-        - os: macos-latest
-          python-version: '3.10'
-          toxenv: py310
-          cache_cmd: -- --dbase-download-dir ~/.chianti
-        # Test the oldest and newest configuration on Windows
-        - os: windows-latest
-          python-version: 3.8
-          toxenv: py38
-          cache_cmd: -- --dbase-download-dir ~/.chianti
-        - os: windows-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/run-tests.yml
```diff
@@ -6,72 +6,26 @@ on:
...
-          python-version: 3.9
-          toxenv: py39
-          cache_cmd: -- --dbase-download-dir ~/.chianti
-        - os: ubuntu-latest
+jobs:
+  test:
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@v1
+    with:
+      posargs: '--dbase-download-dir ~/.chianti'
+      toxdeps: "'tox<4' tox-pypi-filter"
```
Cadair, line 15: You probs don't need this

.github/workflows/run-tests.yml
```diff
@@ -6,72 +6,40 @@ on:
...
-        - os: ubuntu-latest
-          python-version: 3.9
-          toxenv: py39
-          cache_cmd: -- --dbase-download-dir ~/.chianti
-        - os: ubuntu-latest
+        - linux: pre-commit
+          posargs: ''
+      coverage: 'codecov'
+      cache-path: ~/.chianti
+      cache-key: chianti-${{ github.run_id }}
```
Cadair, line 31: What scope is this caching over?

.github/workflows/run-tests.yml
wtbarnes: Oh I wasn't sure if this was fixed upstream/in the template

.github/workflows/run-tests.yml
wtbarnes: Every env can use the same database  so I don't want to download for every env.

My old cache key was this: https://github.com/wtbarnes/fiasco/blob/main/.github/workflows/run-tests.yml#L67. But I wasn't sure how to reproduce that here.

PR #223, opened 2023-02-24
Correct level populations for ionization and recombination processes

### LINKED ISSUE #24: Add correction for ionization and recombination in level populations calculation
There should be a correction for ionization and recombination processes in the level population calculation. Currently, this is not included. See [section 2.3 of Landi et al. (2006)](http://adsabs.harvard.edu/abs/2006ApJS..162..261L) for more details as well as [section 6 of Dere et al. (2009)](http://adsabs.harvard.edu/abs/2009A%26A...498..915D).

wtbarnes:
Fixes #24 

This PR accomplishes two main things:

1. It implements the correction to the level populations calculation by incorporating processes due to level-resolved ionization and recombination as stored in the `cilvl` and `reclvl` files as described in Landi et al. (2006)
2. It fixes the interpolation of the ionization equilibrium by switching the interpolation from linear to a Piecewise Cubic Hermite Interpolating Polynomial (PCHIP) approach. This is more accurate over the full temperature range and even for small ionization fractions.

A few remaining TODOs:

- [x] Tests that compare level populations with and without correction factor (note that this may involve adding another file to the test dataset as few actually have cilvl and reclvl files). This should test that only the levels inlcuded in those files are modified in the level populations calculation.
- [x] Deal with divide-by-zero warning in numerator calculation in correction factor
- [x] Docstring for level populations
- [x] Clarifying comments where needed regarding interpolation methods
- [x] Run IDL comparison checks against continuum results to ensure ioneq interpolation has not caused a regression
- [x] Compare against Fe XX correction factor from v8 IDL code

### DIFF
--- a/fiasco/conftest.py
+++ b/fiasco/conftest.py
@@ -84,6 +84,11 @@ TEST_FILES = {
     'fe_27.rrparams':                       '75383b0f1b167f862cfd26bbadd2a029',
     'fe_10.psplups':                        'dd34363f6daa81dbf106fbeb211b457d',
     'fe_10.elvlc':                          'f221d4c7167336556d57378ac368afc1',
+    'fe_20.elvlc':                          'bbddcf958dd41311ea24bf177c2b62de',
+    'fe_20.wgfa':                           'c991c30b98b03c9152ba5a2c71877149',
+    'fe_20.scups':                          'f0e375cad2ec8296efb2abcb8f02705e',
+    'fe_20.cilvl':                          'b71833c51a03c7073f1657ce60afcdbb',
+    'fe_20.reclvl':                         'cf28869709acef521fb6a1c9a2b59530',
 }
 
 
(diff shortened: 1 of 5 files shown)


wtbarnes:
Based on the test failures, it looks like the free-free and free-bound values have changed, likely due to the differences in how I'm interpolating the ioneq. The differences are surprisingly large which is worrying.

wtbarnes:
For Mg 12, when comparing to the equivalent IDL correction factors sent by Peter, the calculation here shows good agreement,

image

image

wtbarnes:
I've finally settled on using [`PchipInterpolator`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.PchipInterpolator.html#scipy.interpolate.PchipInterpolator) to perform the cubic interpolation of the ionization fractions in log-space. This seems to be best solution for balancing smoothness and reducing oscillations. see plots of Fe V, VI, and XX ionization fractions below:

image

image

wtbarnes:
Comparing the correction factors as implemented here and those in CHIANTI IDL for Fe XX for T=10 MK and $n_e=10^{10}$ $\mathrm{cm}^{-3}$,

image

PR #252, opened 2024-01-12
Fix the definition of hydrogenic and helium-like ions

### LINKED ISSUE #251: Hydrogen- and helium-like atoms only for Z > 6 and 10?
The definition of hydrogen-like requires `Z > 6`:
```return (self.atomic_number - self.charge_state == 1) and (self.atomic_number >= 6)```
and helium-like `Z > 10`:
```return (self.atomic_number - self.charge_state == 2) and (self.atomic_number >= 10)```

Why is this?  Is He II not hydrogen-like?  Or is there some logic or functionality that only applies to elements with higher atomic numbers?

### PR DESCRIPTION
**Pull request by jwreep**
Fixes #251 

Please review.  

I don't think that this line
https://github.com/wtbarnes/fiasco/blob/282b819f5a9b257db93d1d484fa63ada2cb0c610/fiasco/ions.py#L975
needs to be changed since the function it's embedded in is only called for heavier elements.

### DIFF
--- a/fiasco/ions.py
+++ b/fiasco/ions.py
@@ -907,8 +907,8 @@ Using Datasets:
 
         The direction ionization cross-section is calculated one of two ways.
         See :cite:t:`dere_ionization_2007`, Sections 3.1 and 3.2 for details.
-        For H-like ions with :math:`Z\ge6` and He-like ions with :math:`Z\ge10`, 
-        the cross-section is computed according to the method of 
+        For H-like ions with :math:`Z\ge6` and He-like ions with :math:`Z\ge10`,
+        the cross-section is computed according to the method of
         :cite:t:`fontes_fully_1999`,
 
         .. math::
@@ -942,7 +942,7 @@ Using Datasets:
         :math:`U,f,\Sigma` are all stored in the CHIANTI database such that :math:`\sigma_I`
         can be computed for a given :math:`E`.
         """
-        if ((self.hydrogenic and self.atomic_number >= 6) or 
+        if ((self.hydrogenic and self.atomic_number >= 6) or
             (self.helium_like and self.atomic_number >= 10)):
             return self._fontes_cross_section(energy)
         else:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: fiasco/ions.py [hunk 1]
```diff
@@ -940,7 +942,8 @@ def direct_ionization_cross_section(self, energy: u.erg) -> u.cm**2:
         :math:`U,f,\Sigma` are all stored in the CHIANTI database such that :math:`\sigma_I`
         can be computed for a given :math:`E`.
         """
-        if self.hydrogenic or self.helium_like:
+        if (self.hydrogenic and self.atomic_number >= 6) or
+            (self.helium_like and self.atomic_number >= 10):
```
wtbarnes, line 946:
```suggestion
        if ((self.hydrogenic and self.atomic_number >= 6) or
            (self.helium_like and self.atomic_number >= 10)):
```

jwreep: Yeah, just noticed the tests failed.  Fixing it now...

jwreep: Too much C++ in my brain.

jwreep:
I'm not sure I understand that last check that's failing.  Any idea what the issue is?

wtbarnes:
The pre-commit job runs several style checks on the changes to ensure consistent code style across the package. I can fix that easily. You can also install the pre-commit hooks locally using these instructions: https://docs.sunpy.org/en/latest/dev_guide/contents/code_standards.html#formatting and it will fix the files for you.

jwreep:
Thank you!  I didn't know about that.  I'll make sure to use it going forward.

wtbarnes:
@pryoung also pointed out to me that the Fontes and Dere papers note that the fits to the Fontes cross-sections should be accurate for $Z\ge4$, but then there's an additional note that Li III and B V are better fit with the BT scaled cross-section data and that the Be IV cross-sections are fit using the B V cross-sections appropriately scaled by the ionization potential. Thus, the $Z\ge6$ requirement is presumably to ensure those three ions are not fit using Fontes and instead use the `.diparams` values.

Interestingly, `.diparams` values are available for C VI, despite the fact that that cross-section is fit using Fontes. A plot of these two cross-sections is shown below,

image

PR #254, opened 2024-01-12
Implement two-ion model for >= v9 compatibility

### LINKED ISSUE #224: Refactor high-level interface for transitions and levels
There are currently classes for energy levels (`Level`) as well as a class to hold all transitions (`Transitions`). However, they aren't terribly useful and also do not interact with one another.

One possible way to refactor this would be to create a `Transition` class and have `Transitions` hold many instances of `Transition` instances. Similarly, each transition could emit a `Level` class for the upper and lower levels that it represents a transition between. The `Transitions` class would hold many of the same properties that it currently does, but would derive them from these higher-level classes rather than directly from the underlying data itself.

There may be some performance issues with this approach, e.g. every time we want to list all transition wavelengths, it would not make sense to have to rebuild this `Transitions` object from scratch.

### LINKED ISSUE #234: Include collisional-radiative method in level populations calculation
CHIANTI version 9 introudced the `.auto` and `.rrllvl`such that the processes of autoionization and dielectronic recombination from the lower and upper stages, respectively, could be included the level populations calculation. This modification is described in detail in the [Appendix of the CHIANTI v9 paper](https://ui.adsabs.harvard.edu/abs/2019ApJS..241...22D/abstract). 

The `Ion.level_populations` method will need to be modified significantly to incorporate these additional processes that factor into the level populations of the ion.

Before this can be addressed, #48 and #114 will need to be closed such that the needed data can be accessed in order to perform these calculations.

### LINKED ISSUE #344: Separate "A" values in `Transitions` object by whether they are two-photon rates, radiative decay rates, or autoionization rates
In the `wgfa` CHIANTI files, the $A$ value column can correspond to one of three similar but distinct physical quantities:

1. Radiative decay rate between two bound levels
2. Autoionization rate (does not produce a photon)
3. Two-photon rate (produces continuum radiation

The latter two are denoted by zero wavelength in the files. Currently, there is a flag in `Transitions` as to whether the transition corresponds to a two-photon transition or not, but the A values are not separated. No attempt is currently made to separate out the autoionization rates (though there is some work being done in #254 to address this).

Currently, my thinking is that the autoionization and two-photon rates should be entirely separated from the A values exposed by this structure. It is very confusing that they are all grouped together. However, I'm not sure the best way to then expose these values.

For more info, see [`read_wgfa_str.pro`](https://github.com/chianti-atomic/chianti-idl/blob/0387392a6cba39ce700be0d164f53ab56b2fcd12/read/read_wgfa_str.pro)

wtbarnes:
Fixes #234
Fixes #224
Fixes #344

This PR represents the major changes needed for compatibility with versions 9 and later of the CHIANTI database. Most notably, it modifies `Ion.level_populations` method to include processes that couple a given ion and the next highest ionization state through ionization and recombination processes. In addition, this PR also includes the following changes:

## High-level Summary

- Include ionization, autoionization, radiative recombination, dielectronic recombination, and dielectronic capture in the level populations equation.
- Refactor how the rate matrices are compute for all processes used to solve the level populations equation. This includes refactoring and greatly simplifying `Ion.level_populations`
- Add `n_levels` and `n_transitions` property to `Ion`
- Add `is_autoionization` and `is_bound_bound` to `Transitions` for more easily differentiating between different kinds of transitions.
- Rename `Level` to `Levels` and add ability to hold information about multiple energy levels at once.
- Refactor `Transitions` to be more efficient and useful as a high-level object
- Fix a bug in `burgess_tully_descale` to never return object-type arrays
- The default version of the database to be used is now v9.0.1 and v9.0.1 is now in the supported version list
- Add IDL comparison tests for the matrices that connect ionization states in the level population equations

## Checklist

- [x] Add `n_levels` property
- [x] Add `n_transitions` property
- [x] Fix bug in `burgess_tully_descale` to never return object-type arrays
- [x] Review how indexing into rate matrices is done (e.g. vectorize where)
- [x] Separate out code for building coefficient matrices
- [x] Implement ionization rates in coefficient matrix
- [x] Implement level-resolved radiative recombination
- [x] Implement autoionization rates in coefficient matrix
- [x] Dielectronic capture
- [x] Dielectronic recombination
- [x] Refactor how depopulation terms are added along the diagonal
- ~~Logic for excluding ionization/recombination correction factor when two-ion model is used~~
- [x] Logic for turning on/off two-ion model
- [x] Decide where depopulating rates along the diagonal should be added to the rate matrices (in IDl these aren't added until right before solving the rate matrix equation, but in `fiasco` we treat these as terms in the coefficient matrix)
- [x] Tests for two-ion model
-  ~~Docs~~ (see #349)
- [x] Change default version of database to be v9 everywhere
- [x] Add v9 to list of allowed database versions
- [x] IDL comparison tests for two-ion model
- [x] ~~Review where `_elvlc` is used explicitly (favor `levels`)~~ Create an issue to replace all usage of `_elvlc` with `self.levels` (or indexing)
- [x] Create an issue to replace all usage of `_wgfa` with `self.transitions`
- [x] Write a high-level summary of the changes in this PR description
- [x] Add necessary files to test database so that two-ion model is actually run
- [x] Review example gallery
- [x] Update message about database compatibility
- [x] Review test coverage changes.


wtbarnes:
The v8 test failures are just due to a naming change in the variables used in the test data. This is already tracked by #347.

jwreep:
Hooray!  Thanks for your efforts, Will!  That's a major milestone for the package.

PR #259, opened 2024-01-20
Revamp docs and add page on IDL-fiasco comparisons

### LINKED ISSUE #208: Issue with `sphinx-book-theme`
The currently released version of `sphinx-book-theme` is 0.3.3. This release pins to v0.8.1 of `pydata-sphinx-theme` which has a pin to `sphinx>=3.5.4,<5` which for some reason causes very old versions of sphinx to be installed on both GitHub actions and RTD. This in turn causes doc build failures on Python 3.10. 

One way to solve this would be pinning sphinx above 4.2 (I think?) in the fiasco docs dependencies. Alternatively, we could switch to using the pydata-sphinx-theme directly which has a much more recent release that does not suffer from this pinning issue.

### LINKED ISSUE #215: Add page for fiasco/IDL comparisons
It might be useful to have a page of the docs dedicated to showing comparisons better fiasco and IDL for various quantities, e.g. contribution functions, ionization rates, continuum emission. This could serve as a graphical comparison to make sure there is consistency between the two codes. 

This would obviously not be run each time the docs are built. Instead, the IDL results would be stored in a file and plotted. These files would be per Chianti version and also fiasco version. There would be a script in the repo that generated this data and this would be done by hand for each version combination (maybe as part of the version release).

### LINKED ISSUE #255: Inheritance diagrams in docs are broken
All of the inheritance diagrams for the stable version of the documentation are broken, e.g. https://fiasco.readthedocs.io/en/stable/code_ref/io.html#class-inheritance-diagram. This may be a sphinx-related issue as I'm still pinning to `sphinx<5`.

wtbarnes:
Fixes #255
Fixes #208 
Fixes #215 

Remaining TODOs

- ~~Flesh out the atomic database topic guide more~~
- [x] Add the fiasco-IDL comparisons as a topic guide page
- [x] Add the topic guide page to the landing page as a new panel
- [x] Figure out organization of the panels on the landing page
- [x] Fix the doctest
- [x] Add more test files to the database used on RTD


wtbarnes:
The last thing to do is make sure we have all the needed files for computing the rates (ionization and recombination) in the test file list. Currently, the compute rates are not showing up on the rendered docs because those files are missing from the test database.

PR #275, opened 2024-03-20
Add graphviz as an apt dependency for the readthedocs build.

byrdie:
The [Code Reference page](https://fiasco.readthedocs.io/en/stable/code_ref/fiasco.html) has a broken class inheritance diagram.

This PR adds graphviz as an apt dependency to the readthedocs build to attempt to render that diagram correctly.

Apologies if this has already been fixed and I'm just late to the party.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -10,6 +10,8 @@ build:
   os: ubuntu-22.04
   tools:
     python: "mambaforge-4.10"
+  apt_packages:
+    - graphviz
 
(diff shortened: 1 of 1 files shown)


byrdie:
Ok, nevermind. It seems like this is working in the latest version of the documentation, I just didn't know how to find it.

wtbarnes:
No worries! Sorry, yes I need to do another release so that the latest version of the docs will be the first thing you see when you land on RTD. Thanks for the PR!

PR #282, opened 2024-05-04
Add continuum contributions to `IonCollection.radiative_loss`

### LINKED ISSUE #232: `IonCollection.radiative_loss` should include continuum contributions
The `radiative_loss` function should optionally include the continuum contribution from each ion. This could be done as a final step (outside of the main loop) by just calling the appropriate continuum methods on `IonCollection`. 

The one issue is how to handle the wavelength dependence. Currently, `radiative_loss` does not require a wavelength and just sums over every single transition in the ion. However, the continuum methods all require a wavelength. One option would be to have a default wavelength argument (e.g. 1-5000 Å) and then always only include the line transitions that fall within that wavelength range. This would make computing radiative losses for only particular wavelength ranges much easier.

jwreep:
Fixes #232 

I've added f-f and f-b ~~, and 2-p~~ continuum emission to the radiative loss calculation.  

~~It appears to be working~~, but I'd like to do a bit more testing.  

~~One potential issue: the calculation currently assumes a fixed bin width.  It might be nice to generalize and do the integral properly to allow for non-fixed bins, particularly since it can be integrated over a very large range.~~

As a note to myself (@wtbarnes), the following functions still need docstrings before we merge this:

- [x] `IonCollection.radiative_loss_free_free`
- [x] `IonCollection.radiative_loss_free_bound`
- [x] `IonCollection.radiative_loss_bound_bound`

Alternatively, we could make all of these methods private and only expose the `radiative_loss` method.


jwreep:
Figure_1

The output from all ions looks reasonable, except for `log T > 7`.  There should be an increase with temperature from free-free emission (I thought?).

wtbarnes:
Hmm do you have a comparison of the IDL and Python cases? I definitely agree that you should see the uptick in the losses at high temperatures. I had forgotten that IDL computes the radiative loss contribution from the continuum in a slightly different way. I would think these should be consistent though...

Also, if you just plot the wavelength-averaged ff contribution, does that show an increase at high temperatures?

jwreep:
I haven't had much time to come back to this since I posted.  The CHIANTI User's guide shows the uptick, but that plot is from v6 of the database.  

I don't know if it's the Gaunt factor causing the discrepancy, but I've tried changing the wavelength range and binning without any change.  If there's a programming mistake here, it's a subtle one!

wtbarnes:
I wonder if this is a consequence of not including the lower wavelengths that make the dominant contributions at high temperatures? For example, computing just the free-free contribution and averaging over the wavelength dimension,

```python
import astropy.units as u
import fiasco
import numpy as np
import matplotlib.pyplot as plt
from astropy.visualization import quantity_support

temperature = np.logspace(5,8,100) * u.K
all_ions = fiasco.IonCollection(*[fiasco.Ion(i, temperature) for i in fiasco.list_ions()])
wavelength = np.logspace(-2,2,100) * u.AA
ff_short_wave = all_ions.free_free(wavelength).mean(axis=1)
wavelength = np.logspace(1,2,100) * u.AA
ff_long_wave = all_ions.free_free(wavelength).mean(axis=1)

with quantity_support():
    plt.plot(temperature, ff_short_wave, label=r'$\lambda_{min}=0.01$ Å')
    plt.plot(temperature, ff_long_wave, label=r'$\lambda_{min}=10$ Å')
plt.xscale('log')
plt.yscale('log')
plt.legend()
```

gives

image

jwreep:
I tried to redo the function this morning, and while the f-f does have an uptick, I think the b-b is dominating the total losses.  

Here's how I changed the continuum calculation:
```
        wavelength = np.logspace(-2,4,200) * u.angstrom

        ff = self.free_free(wavelength, **kwargs)
        fb = self.free_bound(wavelength, **kwargs)
        tp = self.two_photon(wavelength, density, **kwargs)

        for i in range(len(density)):
            for j in range(len(self.temperature)):
                rad_loss[j,i] += np.trapz(ff[j,:], wavelength)
                rad_loss[j,i] += np.trapz(fb[j,:], wavelength)
                rad_loss[j,i] += np.trapz(tp[:,j,i], wavelength)
```
and I broke it apart by components as well:
Figure_1

It looks like f-f is still small compared to the total, which might be because b-b is too large or one or all of the continua are too small.  Alternatively, maybe this is right?

wtbarnes:
That plot almost looks like the FF and FB are missing some scaling factor. I'm surprised that they are so far below the bound-bound losses as well as being comparable to the two-photon losses at most temperatures.

jwreep:
I think I figured it out.  The discrepancy is the integration.  Changing the code to ignore the bin width
```
        ff = self.free_free(wavelength, **kwargs).sum(axis=1)*u.angstrom
        fb = self.free_bound(wavelength, **kwargs).sum(axis=1)*u.angstrom
        tp = self.two_photon(wavelength, density, **kwargs).sum(axis=0)*u.angstrom
```
I get:

Figure_1

In IDL [ff_rad_loss.pro](https://sohoftp.nascom.nasa.gov/solarsoft/packages/chianti/idl/continuum/ff_rad_loss.pro), the calculation is 
```
rad_loss[k]=rad_loss[k]+this_abund*this_ioneq[k]*factor*sqrt(t[k])*z^2*gaunt[k]
```
which is just a straight sum (no integration).  

So, I guess the question is why the bin width is ignored?  Shouldn't the total radiation for e.g. free-free be given by something like $R_{ff} = \int_{\lambda} I_{ff}(\lambda) d\lambda$ ?

jwreep:
The f-f emissivity according to technical report 11 is 

Screenshot 2024-06-05 at 1 54 47 PM

[CHIANTI's technical report #9](https://www.chiantidatabase.org/tech_reports/09_rad_loss/chianti_report_09.pdf) addresses the radiative losses.  They neglect two-photon's contributions to the total losses.  The equation for free-free (2) doesn't appear to be a direct integration of the emissivity.

Screenshot 2024-06-05 at 1 48 26 PM

Note the exponential factor has disappeared.  This is perhaps why a direct numerical integration is not producing the correct result.  I have no idea where the exponential went, though!

jwreep:
Tentatively fixed.  Still needs a few unit tests to be added, but the losses compare well to IDL.

Screenshot 2024-06-24 at 11 27 47 PM

This uses the same implementation as the IDL formulation, which is based on Mewe et al 1986 for f-b losses and Sutherland 1998 for f-f.  Two-photon is neglected.

jwreep:
@wtbarnes, this should be good to go!

I've followed the implementation in the IDL version, which doesn't integrate the f-f and f-b functions directly, but implements separate functions for radiative losses.  Two-photon emission is neglected in this. 

I pinned the numpy version < 2 for now.  

One small note: the implementation of the f-b Gaunt factor in IDL uses a polynomial approximation to an infinite sum (Eq. 14 of Mewe et al 1986).  Scipy has the appropriate function for the analytic solution, so I've used that instead.  The plot below compares the approximation to the analytic solution.  It should be a negligible difference but could cause a small discrepancy with IDL's output.  We could make the approximation an option, if wanted.

Screenshot 2024-06-25 at 2 26 34 PM

wtbarnes:
Thanks Jeff! This is great and sorry for being so unresponsive on the review front. I will try to get to this either today or early next week. 

Re: your last comment, I think using the scipy implementation is better. I did something similar for the Gauss-Laguerre quadrature in the case of the ionization rates. 

I'm happy to add an IDL comparison test here so that we can check the output as we do any refactoring. Would you mind pasting here snippet of IDL code you were running when you were comparing against the CHIANTI IDL results?

jwreep:
No worries -- enjoy Tenerife!  I'll add the IDL code when I'm in the office tomorrow.

jwreep:
Using `sun_coronal_2021_chianti.abund`:
```
IDL> ff_rad_loss,temperature,ff 
IDL> fb_rad_loss,temperature,fb      
IDL> rad_loss,temperature,total   
IDL> plot,temperature,total,chars=2,xtit='Temperature',ytit='Loss Rate',/ylog,/xlog,yr=[1e-25,1e-21]
IDL> oplot,temperature,fb,line=1,thick=2     
IDL> oplot,temperature,ff,line=2    
IDL> oplot,temperature,(total-ff-fb),line=3     
```
Screenshot 2024-06-27 at 1 32 44 PM

jwreep:
The calculation is extremely slow, but seems to speed up significantly without the warnings.  Perhaps it would be useful to have a keyword to suppress the warnings (or vice versa)?

wtbarnes:
Apologies again that this has taken so long. I've just added a few tests that compare the IDL and Python outputs and made some minor tweaks to one of the gaunt factor functions as I was getting overflow warnings due to large terms in the exponential at low temperatures. These should now be handled properly.

fiasco/ions.py
```diff
@@ -283,6 +284,25 @@ def formation_temperature(self) -> u.K:
         """
         return self.temperature[np.argmax(self.ioneq)]
 
+    @property
+    def zeta_0(self) -> u.dimensionless_unscaled:
```
wtbarnes, line 288: Is this generally useful or could we just make it private?

fiasco/ions.py
```diff
@@ -1316,6 +1336,35 @@ def free_free(self, wavelength: u.angstrom) -> u.erg * u.cm**3 / u.s / u.angstro
...
+
+        Notes
+        -----
+        The result does not include ionization equilibrium or abundance factors.
+
+        The prefactor :math: `F_{k}` is defined in Equation 19 of :cite:t:`sutherland_accurate_1998`,
+        with a value of 1.42555669e-27 cm$^{5}$ g K$^{-1/2}$ s$^{3}$).
+
+        """
+        prefactor = (16./3**1.5) * np.sqrt(2. * np.pi * const.k_B/(const.hbar**2 * const.m_e**3)) * (const.e.esu**6 / const.c**3)
```
wtbarnes, line 1364: Where does the $16/3^{1.5}$ number come from?

fiasco/ions.py
```diff
@@ -1473,6 +1539,63 @@ def free_bound(self,
...
+        z = self.charge_state
+
+        if z == 0:
+            return u.Quantity(np.zeros(self.temperature.shape) * u.erg * u.cm**3 / u.s)
+        else:
+            recombined = self.previous_ion()
+            if not recombined._has_dataset('fblvl'):
+                return u.Quantity(np.zeros(self.temperature.shape) * u.erg * u.cm**3 / u.s)
+
+            C_ff = 64 * np.pi / 3.0 * np.sqrt(np.pi/6.) * (const.e.esu**6)/(const.c**2 * const.m_e**1.5 * const.k_B**0.5)
```
wtbarnes, line 1581: Where does the $64/3$ factor come from?

fiasco/ions.py
```diff
@@ -283,6 +284,25 @@ def formation_temperature(self) -> u.K:
...
+        difference = self.atomic_number - (self.charge_state + 1)
+        if difference <= 0:
+            max_vacancies = 1
+        elif difference <= 8 and difference > 0:
+            max_vacancies = 9
+        elif difference <= 22 and difference > 8:
+            max_vacancies = 27
+        else:
+            max_vacancies = 55
+        return float(max_vacancies - difference)
```
wtbarnes, line 304: Does this need to be explicitly cast to a float?

fiasco/ions.py
```diff
@@ -1316,6 +1336,35 @@ def free_free(self, wavelength: u.angstrom) -> u.erg * u.cm**3 / u.s / u.angstro
...
+        all wavelengths.  The total losses per unit emission measure is then given by
+        (Equation 18 of :cite:`sutherland_accurate_1998`):
+
+        .. math::
+
+            R_{ff}(T_e) = F_{k} \sqrt{(T_{e})} Z^{2} \langle g_{t,ff}\rangle
+
+        where :math: `T_{e}` is the electron temperature, :math: `\F_{k}` is a constant,
+        :math: `Z` is the ionization stage, and :math: `\langle g_{t,ff}\rangle` is the
+        total (wavelength-averaged) free-free Gaunt factor.
```
wtbarnes, line 1354:
```suggestion
        where :math:`T_{e}` is the electron temperature, :math:`F_{k}` is a constant,
        :math:`Z` is the ionization stage, and :math:`\langle g_{t,ff}\rangle` is the
        total (wavelength-averaged) free-free Gaunt factor.
```

fiasco/ions.py
```diff
@@ -1316,6 +1336,35 @@ def free_free(self, wavelength: u.angstrom) -> u.erg * u.cm**3 / u.s / u.angstro
...
+
+        Notes
+        -----
+        The result does not include ionization equilibrium or abundance factors.
+
+        The prefactor :math: `F_{k}` is defined in Equation 19 of :cite:t:`sutherland_accurate_1998`,
+        with a value of 1.42555669e-27 cm$^{5}$ g K$^{-1/2}$ s$^{3}$).
```
wtbarnes, line 1361:
```suggestion
        The prefactor :math:`F_{k}` is defined in Equation 19 of :cite:t:`sutherland_accurate_1998`,
        with a value of 1.42555669e-27 cm$^{5}$ g K$^{-1/2}$ s$^{3}$).
```

I would move this above the Notes section to be just below the equation.

fiasco/ions.py
```diff
@@ -1473,6 +1539,63 @@ def free_bound(self,
...
+
+        .. math::
+
+            R_{fb} = \frac{C_{ff} k_{B} G_{fb} T^{1/2}}{h c} \exp{\Big(\frac{-h c}{\lambda k_{B} T}\Big)}
+
+        We have dropped the factor of :math:`n_{e}^{2}` here to make the loss rate per unit EM.
+
+        Notes
+        -----
+        The input ion `self` is taken to the recombining ion.
```
wtbarnes, line 1570:
```suggestion
        This ion, for which the free-bound radiative loss is being calculated, is taken to the recombining ion.
        The ion one ionization stage lower is taken to be the recombined ion.
```

fiasco/ions.py
```diff
@@ -1473,6 +1539,63 @@ def free_bound(self,
...
+        """
+        z = self.charge_state
+
```
wtbarnes, line 1573:
```suggestion
```

fiasco/ions.py
```diff
@@ -1473,6 +1539,63 @@ def free_bound(self,
...
+        if z == 0:
+            return u.Quantity(np.zeros(self.temperature.shape) * u.erg * u.cm**3 / u.s)
+        else:
+            recombined = self.previous_ion()
+            if not recombined._has_dataset('fblvl'):
+                return u.Quantity(np.zeros(self.temperature.shape) * u.erg * u.cm**3 / u.s)
+
```
wtbarnes, line 1580:
```suggestion
```

fiasco/ions.py
```diff
@@ -1473,6 +1539,63 @@ def free_bound(self,
...
+            E_fb = np.where(E_obs==0*u.erg, E_th, E_obs)
+
+            wvl_n0 = const.h * const.c / (recombined.ip - E_fb[0])
+            wvl_n1 = (recombined._fblvl['n'][0] + 1)**2 /(const.Ryd * z**2)
+
+            g_fb0 = self._gaunt_factor_free_bound_total(ground_state=True)
+            g_fb1 = self._gaunt_factor_free_bound_total(ground_state=False)
+
+            term1 = g_fb0 * np.exp(-const.h*const.c/(const.k_B * self.temperature * wvl_n0))
+            term2 = g_fb1 * np.exp(-const.h*const.c/(const.k_B * self.temperature * wvl_n1))
```
wtbarnes, line 1595:
```suggestion

            E_obs = recombined._fblvl['E_obs']*const.h*const.c
            E_th = recombined._fblvl['E_th']*const.h*const.c
            E_fb = np.where(E_obs==0*u.erg, E_th, E_obs)
            wvl_n0 = const.h * const.c / (recombined.ip - E_fb[0])
            wvl_n1 = (recombined._fblvl['n'][0] + 1)**2 /(const.Ryd * z**2)
            g_fb0 = self._gaunt_factor_free_bound_total(ground_state=True)
            g_fb1 = self._gaunt_factor_free_bound_total(ground_state=False)
            term1 = g_fb0 * np.exp(-const.h*const.c/(const.k_B * self.temperature * wvl_n0))
            term2 = g_fb1 * np.exp(-const.h*const.c/(const.k_B * self.temperature * wvl_n1))
```

fiasco/ions.py
```diff
@@ -1528,6 +1651,77 @@ def _karzas_cross_section(self, photon_energy: u.erg, ionization_energy: u.erg,
...
+
+        where :math: `\zeta(x)` is the Riemann zeta function.
+
+        However, the second sum is analytic, :math: `\sum_{1}^{n_{0}} n^{-3} = \zeta(3) + \frac{1}{2}\psi^{(2)}(n_{0}+1)`
+        where :math: `\psi^{n}(x)` is the n-th derivative of the digamma function (a.k.a. the polygamma function).
+        So, we can write the full solution as:
+        .. math::
+            f_{1}(Z, n, n_{0}) = \zeta(3) - \sum_{1}^{n_{0}} n^{-3} = - \frac{1}{2}\psi^{(2)}(n_{0}+1)
+
+        The final expression is therefore simplified and more accurate than :cite:t:`mewe_freebound_1986`.
```
wtbarnes, line 1688: Because this function is not public API (which I agree with!), this docstring does not show up in the rendered documentation. Should we maybe move this elsewhere in the docs? This seems like useful information to record somewhere. We could potentially leave this to a later PR and just leave it as is for now.

fiasco/ions.py
```diff
@@ -1528,6 +1651,77 @@ def _karzas_cross_section(self, photon_energy: u.erg, ionization_energy: u.erg,
...
+            else:
+                f_1 = -0.5 * polygamma(2, n_0 + 1)
+                with np.errstate(over='ignore'):
+                    f_2 = 2.0 * f_1 * z**4 * np.exp((prefactor * z**2)/((n_0+1)**2))
+
+            # Large terms in the exponential can lead to infs in the f_2 term. These will be zero'd
+            # out when multiplied by the ionization equilibrium anyway
+            f_2 = np.where(np.isinf(f_2), 0.0, f_2)
+
+            return (prefactor * f_2)
```
wtbarnes, line 1722:
```suggestion
        z = self.charge_state
        if z == 0:
            return u.Quantity(np.zeros(self.temperature.shape))
        else:
            recombined = self.previous_ion()
            if not recombined._has_dataset('fblvl'):
                return u.Quantity(np.zeros(self.temperature.shape))
            Ry = const.h * const.c * const.Ryd
            prefactor = (Ry / (const.k_B * self.temperature)).decompose()
            n_0 = recombined._fblvl['n'][0]
            if ground_state:
                g_n_0 = 0.9 # The ground state Gaunt factor, g_fb(n_0), prescribed by Mewe et al 1986
                z_0 = n_0 * np.sqrt(recombined.ip / Ry).decompose()
                # NOTE: Low temperatures can lead to very large terms in the exponential and in some
                # cases, the exponential term can exceed the maximum number expressible in double precision.
                # These terms are eventually set to zero anyway since the ionization fraction is so small
                # at these temperatures.
                with np.errstate(over='ignore'):
                    f_2 = g_n_0 * self.zeta_0 * (z_0**4 / n_0**5) * np.exp((prefactor * z_0**2)/(n_0**2))
            else:
                f_1 = -0.5 * polygamma(2, n_0 + 1)
                with np.errstate(over='ignore'):
                    f_2 = 2.0 * f_1 * z**4 * np.exp((prefactor * z**2)/((n_0+1)**2))
            # Large terms in the exponential can lead to infs in the f_2 term. These will be zero'd
            # out when multiplied by the ionization equilibrium anyway
            f_2 = np.where(np.isinf(f_2), 0.0, f_2)

            return (prefactor * f_2)
```

fiasco/ions.py
jwreep: [Sutherland 1998](https://ui.adsabs.harvard.edu/abs/1998MNRAS.300..321S/abstract) spells out the whole equation.
Screenshot 2024-08-09 at 12 57 01 PM

fiasco/ions.py
jwreep: Likely could be made private.  It might have some useful fringe cases where it would be useful to have access to, but that would be rare.

fiasco/ions.py
jwreep: This whole function is Equation 1a of [Mewe et al 1986](https://ui.adsabs.harvard.edu/abs/1986A%26AS...65..511M/abstract), but they don't spell out the constants in that paper:

Screenshot 2024-08-09 at 1 04 18 PM
Their earlier paper [Gronenschild & Mewe 1978](https://ui.adsabs.harvard.edu/abs/1978A%26AS...32..283G/abstract) writes out the constants in full (with some unit changes):

Screenshot 2024-08-09 at 1 05 55 PM

fiasco/ions.py
jwreep: Not sure why I did that.  Probably not necessary.

fiasco/ions.py
jwreep: I didn't realize this wouldn't show up.  This absolutely needs to be recorded since we've changed the implementation slightly from Chianti, which uses the polynomial form.  I'm not sure where else to suggest that it be moved to, though.

fiasco/ions.py
wtbarnes: Ah I see ok. As long as we have a reference and we've defined the constant as in the paper I think that's fine.

fiasco/ions.py
wtbarnes: The latest version of the docs has a section called "Topic Guides" that is reserved for more non-code explanatory pages: https://fiasco.readthedocs.io/en/latest/topic_guides/index.html. We could put it there.

fiasco/ions.py
jwreep: I had put the references in the doc string for the function.  If it's not clear, we can always add more detail.

fiasco/ions.py
jwreep: Alright, I've added it now.  Hopefully I've formatted it correctly . . . 

Also, thanks for catching the bug in the test!  Beat me to it.

jwreep:
I've added the doc strings now and some tests to keep Codecov happy.  I would suggest keeping each new collection method public -- sometimes it's helpful to break apart the contributions to different emission processes.

jwreep:
Codecov is still showing a gap for the try-except statements in the f-f and f-b radiative loss collection functions.  I'm not sure if it's possible for either to currently throw an exception.  The f-b has some error handling in the corresponding ion class function (returns zeroes), while the f-f doesn't use any files particular to a given ion.  

So, I think the options are to remove the try-except statements or keep the gap in codecov.  I don't think there's anything wrong with keeping the error handling in case something changes in the future, but it shouldn't matter at present.

wtbarnes:
Don't worry about the code coverage for those parts. I'm not so worried about that "failure".

wtbarnes:
Ok I did some minor fixing of the formatting for the new topic guide page. The equations weren't rendering quite right. Once the docs build preview finishes, give that a look over and let me know if it's good to go. Other than that, once the tests pass, I think this is ready.

jwreep:
The math looks like it rendered fine this time.  It should be ready to merge once it finishes the last few tests, then!

wtbarnes:
Ok I found some more nitpicky formatting fixes, almost all caused by me. I also moved the two-photon continuum explanation to a topic guide as well as I found it a bit overly long for a docstring. I'm declaring this the last round of fixes and actually merging once the tests pass this time!

wtbarnes:
Thanks as always! And sorry that this once again took forever to get merged.

PR #286, opened 2024-05-21
Fixes to documentation theme and package infrastructure fixes

nabobalis:
This PR provides a number of fixes to the documentation, particularly in the appearance of the Pydata theme. There are also a number of fixes to how the formatting tools are applied as well as other infrastructure bits.

eduard-khil_trans_NvBQzQNjv4BqVKOd1ILXaurCHi1o16wQuQanpLbznLmtmLfsIXrFY2I

### DIFF
--- a/fiasco/__init__.py
+++ b/fiasco/__init__.py
@@ -1,11 +1,11 @@
 """
-A Python interface to the CHIANTI atomic database
+fiasco: A Python interface to the CHIANTI atomic database
 """
-from fiasco.collections import *
-from fiasco.elements import *
-from fiasco.fiasco import *
-from fiasco.ions import *
-from fiasco.levels import *
+from fiasco.collections import IonCollection
+from fiasco.elements import Element
+from fiasco.fiasco import list_elements, list_ions, proton_electron_ratio
+from fiasco.ions import Ion
+from fiasco.levels import Level, Transitions
 from fiasco.util.util import setup_paths
 
 try:
@@ -19,3 +19,5 @@ defaults = setup_paths()
 from fiasco.util.logger import _init_log
 
 log = _init_log()
+
+__all__ = ["IonCollection", "Element", "list_elements", "list_ions", "proton_electron_ratio", "Ion", "Level", "Transitions", "defaults", "log", "__version__"]
--- a/fiasco/io/__init__.py
+++ b/fiasco/io/__init__.py
@@ -1,6 +1,9 @@
 """
-I/O utilities for reading and building CHIANTI data files
+Parsers for raw CHIANTI atomic data.
+fiasco uses these parsers to transform the raw ASCII files into an HDF5 version of the CHIANTI database.
 """
-from fiasco.io.datalayer import *
-from fiasco.io.factory import Parser
+from fiasco.io.datalayer import DataIndexer, DataIndexerHDF5
+from fiasco.io.factory import Parser, ParserFactory
 from fiasco.io.generic import GenericIonParser, GenericParser
+
+__all__ = ["DataIndexer", "DataIndexerHDF5", "Parser", "ParserFactory", "GenericIonParser", "GenericParser"]
--- a/fiasco/io/__init__.py
+++ b/fiasco/io/__init__.py
@@ -3,7 +3,7 @@ Parsers for raw CHIANTI atomic data.
 fiasco uses these parsers to transform the raw ASCII files into an HDF5 version of the CHIANTI database.
 """
 from fiasco.io.datalayer import DataIndexer, DataIndexerHDF5
-from fiasco.io.factory import Parser, ParserFactory
+from fiasco.io.factory import Parser
 from fiasco.io.generic import GenericIonParser, GenericParser
 
-__all__ = ["DataIndexer", "DataIndexerHDF5", "Parser", "ParserFactory", "GenericIonParser", "GenericParser"]
+__all__ = ["DataIndexer", "DataIndexerHDF5", "Parser", "GenericIonParser", "GenericParser"]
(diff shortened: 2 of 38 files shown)


fiasco/io/factory.py
```diff
@@ -58,9 +58,6 @@ class Parser(GenericParser, metaclass=ParserFactory):
     The Parser ingests the name of a raw ASCII data file and builds an
     `astropy.table.QTable` from it. A predefined parser is created based
     on the file extension, but a custom parser can also be used.
-
-    Examples
-    --------
     """
     def __init__(self, filename, custom_parser=None, **kwargs):
-        super().__init__(filename, **kwargs)
+        super().__init__(filename, custom_parser=custom_parser, **kwargs)
```
nabobalis, line 63: Is this ok?

fiasco/__init__.py
```diff
@@ -1,11 +1,11 @@
...
-from fiasco.collections import *
-from fiasco.elements import *
-from fiasco.fiasco import *
-from fiasco.ions import *
-from fiasco.levels import *
+from fiasco.collections import IonCollection
+from fiasco.elements import Element
+from fiasco.fiasco import list_elements, list_ions, proton_electron_ratio
+from fiasco.ions import Ion
+from fiasco.levels import Level, Transitions
```
wtbarnes, line 8: Why is this not allowed? This defeats the purpose of being able to use `__all__` to define what the top level API should be for each of these modules.

fiasco/__init__.py
```diff
@@ -19,3 +19,5 @@
 from fiasco.util.logger import _init_log
 
 log = _init_log()
+
+__all__ = ["IonCollection", "Element", "list_elements", "list_ions", "proton_electron_ratio", "Ion", "Level", "Transitions", "defaults", "log", "__version__"]
```
wtbarnes, line 23: Same as above. Why.

fiasco/__init__.py
nabobalis: This is allowed, I just wanted to make it super explicit. I can undo this.

fiasco/io/factory.py
wtbarnes: Ok so what is happening is that at the factory level, `custom_parser` is being pulled from the `kwargs`. Instead of adding this here, the explicit keyword argument should be removed from `Parser.__init__`.

fiasco/__init__.py
wtbarnes: Ok I think I understand the logic here. I do like the idea of more explicitly defining the top level namespace here.

fiasco/io/__init__.py
```diff
@@ -1,6 +1,9 @@
 """
-I/O utilities for reading and building CHIANTI data files
+Parsers for raw CHIANTI atomic data.
+fiasco uses these parsers to transform the raw ASCII files into an HDF5 version of the CHIANTI database.
 """
-from fiasco.io.datalayer import *
-from fiasco.io.factory import Parser
+from fiasco.io.datalayer import DataIndexer, DataIndexerHDF5
+from fiasco.io.factory import Parser, ParserFactory
```
wtbarnes, line 6: The factory should not be in the top level namespace

fiasco/io/__init__.py
```diff
@@ -1,6 +1,9 @@
...
 from fiasco.io.generic import GenericIonParser, GenericParser
+
+__all__ = ["DataIndexer", "DataIndexerHDF5", "Parser", "ParserFactory", "GenericIonParser", "GenericParser"]
```
wtbarnes, line 9:
```suggestion
__all__ = ["DataIndexer", "DataIndexerHDF5", "Parser", "GenericIonParser", "GenericParser"]
```

fiasco/io/factory.py
```diff
@@ -1,11 +1,12 @@
 """
-Factory and interface for file parser
+Factory and interface for file parser.
 """
 import pathlib
 
 from fiasco.io.generic import GenericParser
 from fiasco.io.sources import *
 
+__all__ = ['Parser', 'ParserFactory']
```
wtbarnes, line 9:
```suggestion
__all__ = ['Parser']
```

fiasco/util/exceptions.py
```diff
@@ -1,7 +1,9 @@
 """
-Custom exceptions
+Custom exceptions for fiasco.
 """
 
+__all__ = ['MissingIonError', 'MissingDatabaseError', 'MissingASCIIFileError',
+              'MissingDatasetException', 'UnsupportedVersionError']
```
wtbarnes, line 6:
```suggestion
__all__ = [
    'MissingIonError', 
    'MissingDatabaseError', 
    'MissingASCIIFileError',
    'MissingDatasetException', 
    'UnsupportedVersionError'
]
```

The indentation here was weird. Seems sensible to put these all on separate lines?

nabobalis: Yeah.

fiasco/util/setup_db.py
```diff
@@ -28,7 +28,7 @@
 ]
 LATEST_VERSION = SUPPORTED_VERSIONS[-1]
 
-__all__ = ['check_database', 'download_dbase', 'get_test_file_list', 'build_hdf5_dbase']
+__all__ = ['check_database', 'check_database_version', 'download_dbase', 'md5hash', 'get_test_file_list', 'build_hdf5_dbase']
```
wtbarnes, line 31: I think I should just move these functions to be private. I excluded from the top level namespace here for a reason.

wtbarnes: This can be separate PR.

PR #297, opened 2024-08-31
Make two-photon result shape consistent and add IDL comparison test

### LINKED ISSUE #277: Add an IDL comparison test for the two-photon continuum
#260 added the ability to calculate the two photon continuum emission. This included several unit tests, but still needs a comparison test against the IDL code.

### LINKED ISSUE #296: Add labels for running different versions of the tests
The full test suite on the CI currently takes a very long time to run because of all of the different versions of the database we test against. We should add a conditional that only runs these parts of the tests on release or when merging into a release branch or when a particular tag is added to that PR.

wtbarnes:
Fixes #277 
Fixes #296 

This PR does three things:
- Fixes the shape of the two photon result such that it is ordered `(n_temperature, n_density, n_wavelength)` consistent with all of the other continuum methods as well as the level populations method
- Adds a test comparing the two-photon continuum result from IDL and that from fiasco
- Adds a two-photon comparison section to the example gallery 

After adding the comparison test, it's clear that there is some offset between the IDL and Python results. In particular, plotting them side-by-side:

image

The previously written tests all pass so I'm quite sure this isn't related to any changes I've made here.

@jwreep do you know what could be happening here? Is there some factor I'm missing when doing the comparison between the two?


jwreep:
I'm not sure off the top of my head.  Is the normalized difference (IDL - fiasco) / IDL ?  I think the IDL function had a factor of $4 \pi$ that fiasco didn't (sr units) -- is that the discrepancy?  

Presumably this is without protons included?

wtbarnes:
> I'm not sure off the top of my head. Is the normalized difference (IDL - fiasco) / IDL ? 

Yes

> I think the IDL function had a factor of 4 π that fiasco didn't (sr units) -- is that the discrepancy?

I think I'm accounting for this by multiplying the IDL results by a factor of 4pi first before comparing. 

> Presumably this is without protons included?

Correct

wtbarnes:
Note to myself: I need to revert the way the scaling is being applied to the IDL results. The free-free continuum shouldn't be multiplied by 4 pi but the free-bound and two-photon results should be. Despite what the IDL code says, I don't think the free-free code actually gives the result in units of per steradian though I should verify this.

wtbarnes:
Ok actually dealing with the factor of 4 pi things look much better, but there is still something odd going on:

image

Most of the normalized differences are still under 2% and the largest differences seem to be where the continuum emission is quite low so I'm not so worried about those. There is however quite a bit of structure to these differences. In particular, the biggest, most concerning differences seem to be near 300 Å at lower temperatures.

I'm also not actually sure why the structure of the two-photon continuum changes so much around 300 Å for all temperatures.

jwreep:
Not sure why!  The same discrepancy was in the plots I had made in the original pull request.  The bump at 304 is He II's contribution to the emission, but He II wasn't missing any input data, nor did it have a dielectronic ion.

wtbarnes:
> Not sure why! The same discrepancy was in the plots I had made in the original pull request. The bump at 304 is He II's contribution to the emission, but He II wasn't missing any input data, nor did it have a dielectronic ion.

Huh I wonder why fiasco is seemingly under-predicting the He II 304 emission...

jwreep:
A few possible avenues to check:

-- The streaks in the (fiasco-IDL) plots, near the bottom.  The lowest wavelengths for O VIII, C VI, Fe XXVI are $\approx$ 34, 19, 2 Angstrom.  I wonder if there's an issue with all hydrogenic ions, not just He II.  Is there a discrepancy for hydrogen (check 1215 A and up)?

-- Or, can we determine whether helium-like ions have an issue?  Could you zoom in on the region below around 40 A?  O VII, for example, should have a minimum near 21.8 A.

-- IDL uses [a cubic spline fit](https://www.nv5geospatialsoftware.com/docs/SPL_INTERP.html), but it might be worth verifying that it gives the same as fiasco (for the $y$ vs $\psi(y)$ values).

-- The `Z_0` value read in by the H-sequence parser is not used, nor do I see it used in the IDL.  I don't understand its label as "nominal atomic value".

jwreep:
Alright, I checked the spline fits, and there is a difference in what IDL and Python give.  It seems like this might be (one of?) the sources of errors.  Here's a comparison calculated for He II from 300 to 600 Angstrom:
spline_ratio

I'm not sure whether there are two equivalent spline fits that could be used to ensure better agreement.

wtbarnes:
Hmm that's interesting and it would certainly explain the large differences we see near those wavelengths. What does the uninterpolated data look like on top of those curves? I mainly ask because I've very recently found cases where the IDL interpolation is getting it quite wrong especially near edges/discontinuous features in the original data.

wtbarnes:
I did a bit more experimentation with this. Here's a comparison of He II showing the IDL interpolated values, the Python interpolation, as well as the original data and the relative difference (on the right axis).

image

I experimented with a few different interpolation schemes on the Python side and they all seem to show at least a 10% near $\lambda_0$. The `PChipInterpolator` (shown here) seems to do the best, but again still shows a 10% difference. 

It's entirely possible the IDL interpolation is not an appropriate standard to comare against, but looking at the two results compared to the data, it is hard to say which is better (IDL or Python). There are no obvious artifacts in either (e.g. ringing or overshooting).

wtbarnes:
Ok if I use the IDL interpolation scheme and leave everything else the same, I get the following:

image

So the interpolation is certainly the culprit in terms of the differences at 300 Å as well as at the highest temperatures, but its not the only thing going on. However, it does push the relative differences below 10% nearly everywhere (except at the lowest temperatures which is I suspect due to numerical differences in the level populations there).

jwreep:
> but its not the only thing going on. 

The vertical streaks look like ion formation ranges (or EM loci) to me.  Are these the formation temperatures for dielectronic ions that are missing from fiasco?

wtbarnes:
Interesting. I made a comparison with the "clamped" option and found that the differences compared to what IDL is doing are even larger. Let me go back and check again and remake the above plot…maybe I did something wrong.

wtbarnes:
> > but its not the only thing going on.
> 
> The vertical streaks look like ion formation ranges (or EM loci) to me. Are these the formation temperatures for dielectronic ions that are missing from fiasco?

That's certainly possible.

wtbarnes:
Here's the `bc_type="clamped"` result.

image

wtbarnes:
OK, here's the updated comparison over a wider wavelength and lower temperature range

image

I think the differences here are due to differences in the level populations calculation and all seem to be order 10% or less. As you point out, this may be due to the lack of dielectronic ions here as well. I'm inclined to document this as a known issue and call this PR done.

jwreep:
I think that's reasonable.  Sorry for not catching the interpolation issue on the initial pull request -- should've found that a while back.  I had assumed that it was primarily an atomic data issue.  

I don't know what other possible sources of error might exist, so I agree it would be best to revisit this once dielectronic recombination is implemented into the level population calculation.

wtbarnes:
No worries. These interpolation problems are all over the place and seem to manifest themselves in different ways every time. My main objective has always just been to document why we choose particular interpolation schemes or inputs.

PR #316, opened 2024-09-20
Generalize `needs_dataset` decorator to non-Ion objects

### LINKED ISSUE #315: Generalize the decorator @needs_dataset, which assumes that the input is a fiasco.Ion
While testing the Gaunt factor objects in #293, I'm running into missing datasets because of the database version.  While doing so, I noticed that the decorator `@needs_dataset` assumes that the input is a `fiasco.Ion` object, which results in errors similar to this: 

`FAILED fiasco/tests/test_gaunt.py::test_gaunt_factor_free_free_total_itoh[3-True-False-75-1.34662286] - AttributeError: 'GauntFactor' object has no attribute 'ion_name'`

The real error is that it hasn't found `itoh_integrated_gaunt_nonrel.txt` when parsing in the data files, so this is a bit misleading.  

Anyway, just a note that this decorator ought to be generalized, or a new decorator should be created for other types of objects.

wtbarnes:
Fixes #315

### DIFF
--- a/fiasco/util/decorators.py
+++ b/fiasco/util/decorators.py
@@ -17,16 +17,18 @@ def needs_dataset(*names):
 
     def decorator(func):
         """
-        func is a method of `fiasco.Ion`.
+        func is a method of an object that requires a dataset, e.g Ion.
         """
         @wraps(func)
         def func_wrapper(*args, **kwargs):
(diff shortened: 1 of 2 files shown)


jwreep:
That was fast!  Thanks!

PR #329, opened 2024-10-01
Add how-to guide for rebuilding the HDF5 database

### LINKED ISSUE #261: Add a how-to guide in the docs showing how to rebuild of database
While working on #260, I had updated the units for the `hseq_2photon` parser (to remove units from `A_sum`).  The database then requires a rebuild in order to correctly parse the updated units, since it has already hard-coded them.  

One thought: any update to fiasco that changes parser information like this should then require a rebuild of the database to prevent similar issues.  Is there any way to do so?

jwreep:
Fixes #261 

(Hopefully.  New to Sphinx.)


wtbarnes:
Thanks for this! I wonder if instead of having a combination of shell command and `check_database` we could just recommend using `build_hdf5_dbase` with the modification to overwrite the file introduced in #332. We could still recommend passing in the parameters from `fiasco.defaults` for both the ASCII and HDF5 database.

jwreep:
Oh, nice!  That's definitely better.  There was one issue in the back of my head: `rm` doesn't work on Windows, so as I've currently written this, it's only for Unix-based systems.

wtbarnes:
Great! Yeah the OS-independent issue was why I'm in favor of avoiding the shell commands.

 I just merged that other PR so if you rebase (or merge in main), then we should be able to tweak this to just use `build_hdf5_dbase` with the `overwrite=True` keyword argument.

jwreep:
Thanks, I've updated it now.  Passed pytest, but sometimes Github catches issues that the command line seems to miss.  Hopefully it compiles without issue.

docs/how_to_guides.rst
```diff
@@ -65,6 +65,23 @@ For example, a user who wants to place the HDF5 file in a custom folder would pl
...
+-----------------------------------------------
+
+There are occasional changes to fiasco that require the HDF5 database to be rebuilt because of changes to the data structure, such as when there are
+updates to the data parsing code, or when additional data files have been added to the database.  In these cases, the HDF5 database can be
+rebuilt by running the following code in python:
+
+.. code-block:: pycon
+
+   >>> import fiasco
+   >>> fiasco.util.build_hdf5_dbase(fiasco.defaults['ascii_dbase_root'], fiasco.defaults['hdf5_dbase_root'], overwrite=True)  # doctest: +SKIP
```
wtbarnes, line 80:
```suggestion
   >>> fiasco.util.build_hdf5_dbase(fiasco.defaults['ascii_dbase_root'], 
                                    fiasco.defaults['hdf5_dbase_root'], 
                                    overwrite=True)  # doctest: +SKIP
```

wtbarnes (approved):
Minor formatting suggestion for readability in rendered docs. Otherwise looks good to me!

jwreep:
Alright, dumb question -- how do you squash the merge for future reference?  The merge looks like it kept the last commit title rather than the pull request title.  (Sorry about that!).  I don't think I saw any option besides "Merge"

wtbarnes:
No worries. I'm not super worried about keeping the history absolutely pristine. The PR was relatively short anyways so it didn't add too many commits.

When you go to the merge the PR, the big green button should have drop down that allows you to select whether you want to do a merge commit, a squash merge, or a rebase. I've now configured the repo such that only squash merges are allowed so this shouldn't be problem in the future!

jwreep:
Haha, thanks!  My bad, again.  It's not necessarily about being "tidy" but it is helpful if the latest update to a file is not simply labeled "pre-commit" so we can tell what the changes are.

wtbarnes:
Ha fair point. It also means the commit history of any one PR can be as messy as we want and as long as that PR passes, then each commit in the history of the repo will constitute a working version of the code.

PR #337, opened 2024-12-05
Replace `ioneq` with `ionization_fraction` and add setter

### LINKED ISSUE #324: Rename `ioneq` to `ionization_fraction`
"ioneq" is typically used in CHIANTI to denote the ionization fraction in equilibrium and is used as the name of the property on the `Ion` object to represent this quantity. However, I find the name quite cryptic and not overly descriptive. This should be changed to `ionization_fraction`.  This is straightforward to do, but should be done carefully.

### LINKED ISSUE #326: Add setter for ionization fraction
Similar to #325 and how the abundance is handled, there should be logic for passing in an array of ionization fractions in lieu of using the predefined dataset. By default, this would always just use those tabulated values provided by CHIANTI. There would also need to be a check that this array has the same shape as the temperature array, but this could be done in the setter. 

This should be done in tandem with #324

jwreep:
Fixes #324 
Fixes #326 
Also fixes a minor bug introduced by #332 that was breaking `generate_hash_tables` because the path to the hash tables changed.  

- [x] Remove lingering ._ioneq -- not sure what needs to change with the parser but I couldn't get this to work
- [x] Add setter
- [x] Tests for setter


jwreep:
@wtbarnes, how were the `.asdf` files with the IDL data generated?  They have `ioneq` keys in their data, so fully purging the repository of references to `ioneq` will require updating them.  Alternatively, we can just leave those keys in the sections with IDL comparisons, which looks like it is only user-facing in the Docs examples.

wtbarnes:
> @wtbarnes, how were the `.asdf` files with the IDL data generated? They have `ioneq` keys in their data, so fully purging the repository of references to `ioneq` will require updating them. Alternatively, we can just leave those keys in the sections with IDL comparisons, which looks like it is only user-facing in the Docs examples.

The `.asdf` files containing the IDL results are automatically generated by the tests if the files do not exist already: https://github.com/wtbarnes/fiasco/blob/b74ecc066e8cb8ed1221e5469a7c2c27133959b8/fiasco/tests/idl/helpers.py#L127-L129. 

That being said, I do not think it is necessary to remove the usage of "ioneq" when it is in the context of the IDL results as that is what the quantity is called in the IDL code. I would actually prefer to preserve that nomenclature in that context.

As such, any reference to "ioneq" in the tests when referring to the IDL results can be left as is. All of the test files can also remain unchanged, both in name and their contents.

fiasco/util/util.py
```diff
@@ -113,8 +113,9 @@ def walk_sub_dir(subdir):
 
         return subdir_files
 
-    non_ion_subdirs = ['abundance', 'ioneq', 'ip', 'continuum', 'dem']
+    non_ion_subdirs = ['abundance', 'ip', 'continuum', 'dem']
```
wtbarnes, line 116: This should not be removed.

tools/generate_hash_table.py
```diff
@@ -35,7 +35,7 @@ def build_hash_table(dbase_root):
 @click.command()
 @click.option('--database', required=True, type=str)
 def write_hash_table(database):
-    data_dir = pathlib.Path(get_pkg_data_path('data', package='fiasco.util'))
+    data_dir = pathlib.Path(get_pkg_data_path('data', package='fiasco.tests'))
```
wtbarnes, line 38: Wow, I don't know how that happened. Thanks for catching that!

fiasco/ions.py
```diff
@@ -40,7 +40,7 @@ class Ion(IonBase, ContinuumBase):
         input formats.
     temperature : `~astropy.units.Quantity`
         Temperature array over which to evaluate temperature dependent quantities.
-    ioneq_filename : `str`, optional
+    ionization_fraction : `str`, optional
         Ionization equilibrium dataset
```
wtbarnes, line 44:
```suggestion
    ionization_fraction : `str` or `float` or array-like, optional
        If a string is provided, use the appropriate "ioneq" dataset. If an array is provided, it must be the same shape
        as ``temperature``. If a scalar value is passed in, the ionization fraction is assumed constant at all temperatures.
```

wtbarnes (changes requested):
A few suggestions regarding how the setter is handled and some docstring modifications. Thanks as always for taking this on.

This is a breaking change to the API, but I'm not so worried about that at this stage.

tools/generate_hash_table.py
jwreep: I only noticed by chance.  I was trying to build the hash files as a test and it spit out an error.  Simple enough fix!

fiasco/base.py
```diff
@@ -152,7 +157,10 @@ def property_template(self):
         return DataIndexer.create_indexer(self.hdf5_dbase_root, data_path)
 
     property_template.__doc__ = f'Data in {filetype} type file'
-    property_template.__name__ = f'_{"_".join(filetype.split("/"))}'
+    if property_name:
+        property_template.__name__ = f'_{"_".join(property_name.split("/"))}'
+    else:
+        property_template.__name__ = f'_{"_".join(filetype.split("/"))}'
```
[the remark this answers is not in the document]
jwreep: The filetype `.ioneq` and the variable name `ionization_fraction` no longer match, so I was running into issues.  Not sure if this is necessary -- I'll play around with it and see if there's something simpler.

fiasco/ions.py
jwreep: `If a scalar value is passed in, the ionization fraction is assumed constant at all temperatures.`

I hadn't added this functionality.  Does your suggestion above fix this?

fiasco/util/util.py
jwreep: Let me double check this.  I changed it because it was giving errors relating to missing data, presumably because of the mismatch in `.ioneq` and `ionization_fraction` again.  Certainly possible that I'm just being dumb!

fiasco/base.py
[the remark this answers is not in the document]
jwreep: Looks like this isn't necessary.  Vestigial change from my initial commit!  I'll remove it.

fiasco/util/util.py
jwreep: Yup, just being dumb.  I swear there were errors, but I can't reproduce them now, probably because I made a mistake somewhere else that is now fixed.

Oh well, thanks for catching it!

fiasco/ions.py
```diff
@@ -184,8 +191,40 @@ def transitions(self):
...
+            self._ionization_fraction = self.calculate_ionization_fraction()
+        else:
+            ionization_fraction = np.atleast_1d(ionization_fraction)
+            if np.min(ionization_fraction) < 0.0 or np.max(ionization_fraction) > 1.0:
+                raise ValueError("Ionization fractions must be between 0 and 1, inclusive.")
+            elif ionization_fraction.shape != self.temperature.shape:
+                raise ValueError("Ionization fraction array must match the shape of the temperature array.")
+            else:
+                self._dset_names['ionization_fraction'] = None
+                self._ionization_fraction = ionization_fraction
```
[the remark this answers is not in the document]
jwreep: The `ValueError` here is no longer needed with this change.  If an array of incorrect size is passed in, the multiplication raises a `ValueError` already.

jwreep:
This completely fell off my radar.  Been too busy with teaching.  Codecov is complaining that I'm not testing the tests for some reason, but I think it's otherwise good to go.  Let me know if there are any other changes you'd like!

fiasco/util/util.py
```diff
@@ -88,7 +88,7 @@ def get_chianti_catalog(ascii_dbase_root):
     -------
     : `dict`
         All CHIANTI files, separated by category. The resulting dictionary should have the
-        following keys: 'abundance_files', 'ioneq_files', 'ip_files', 'continuum_files',
+        following keys: 'abundance_files', 'ionization_files', 'ip_files', 'continuum_files',
```
wtbarnes, line 91:
```suggestion
```

I think this just wasn't updated after going back to keeping "ioneq" lower down in this function

tools/generate_hash_table.py
```diff
@@ -23,7 +23,7 @@ def build_hash_table(dbase_root):
     filepaths = chain(
         map(build_ion_path, catalogue['ion_files']),
         map(lambda x: pathlib.Path('abundance') / x, catalogue['abundance_files']),
-        map(lambda x: pathlib.Path('ioneq') / x, catalogue['ioneq_files']),
+        map(lambda x: pathlib.Path('ioneq') / x, catalogue['ionization_files']),
```
wtbarnes, line 26:
```suggestion
```

Same as above.

wtbarnes (approved):
Just two small changes that I think got forgotten as part of the last round of corrections. Other than that, everything looks good! Thanks for slogging through this and apologies for the really delayed review.

wtbarnes:
Oh no. I messed this up and accidentally deleted those lines. 🤦 Let me fix that really quick.

wtbarnes (approved):

wtbarnes:
Thanks again Jeff!

jwreep:
Thanks for finishing up the last few bits!  I haven't had enough time to work on fiasco, but I'd like to get back to it soon.

PR #404, opened 2025-08-06
Update versions of pre-commit hooks

namurphy:
The purpose of this PR is to run `pre-commit autoupdate` and update the code accordingly.  There are a few syntax errors found by ruff.


.pre-commit-config.yaml
```diff
@@ -1,12 +1,16 @@
+ci:
+  autofix_prs: false
+  autoupdate_schedule: monthly
```
namurphy, line 3:
```suggestion
  autoupdate_schedule: quarterly
```
...in case you prefer less frequent auto-updates.

PR #407, opened 2025-08-06
Make pytest report the slowest running tests

namurphy:
This PR adds the `--durations` flag to pytest to print out the tests with the slowest durations, and the `--durations-min` flag to set the minimum duration to that would be reported.

> [!CAUTION]
> The `--durations-min` is actually in seconds, despite the ambiguous name.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -27,6 +27,8 @@ commands =
     --cov-report=xml \
     --cov=fiasco \
     {toxinidir}/docs \
+    --durations=10 \
+    --durations-min=0.5 \
     {posargs}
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -27,6 +27,8 @@ commands =
     --cov-report=xml \
     --cov=fiasco \
     {toxinidir}/docs \
+    --durations=10 \
```
namurphy, line 30:
```suggestion
    --durations=20 \
```
...in case you'd like more extensive durations reporting.

PR #411, opened 2025-08-18
Implement dielectronic recombination suppression calculation

### LINKED ISSUE #233: Density-dependent suppression factors for dielectronic recombination rates
Currently, the dielectronic recombination rate calculations in `fiasco` is density-independent though in reality, dielectronic recombination may be suppressed at very high densities. In the IDL routines, there is a correction factor implemented to account for this density sensitivity. I am not sure in which CHIANTI version this correction factor was introduced.

This report gives a brief description of this method and its implementation in CHIANTI: http://www.chiantidatabase.org/tech_reports/17_dr_suppress/chianti_report_17.pdf

wtbarnes:
Fixes #233 

## TODOs

- [x] Reconcile differences with IDL and published result
- [x] Move to `Ion`?
- [x] Implement special cases for $p_{N,0}$ in the lower-temperature correction.
- [x] Tests
- [x] Comparison tests with IDL
- [x] Vectorize over arrays of density?
- ~~Finish docstring~~
- ~~Decide on how this gets integrated into actual DR calculation (could be left to another PR)~~
- ~~More guidance from chianti-atomic/chianti-idl#17~~

### DIFF
--- a/fiasco/tests/test_ion.py
+++ b/fiasco/tests/test_ion.py
@@ -572,8 +572,9 @@ def oxygen(hdf5_dbase_root):
 def test_dielectronic_recombination_suppression_different_sequences(oxygen):
     density = np.logspace(0, 15, 30) * u.cm**(-3)
     for ion in oxygen[1:-1]:
-        suppression = ion._dielectronic_recombination_suppression(density)
-        assert suppression.shape == density.shape
+        suppression = ion._dielectronic_recombination_suppression(density,
+                                                                  couple_density_to_temperature=False)
+        assert suppression.shape == density.shape + oxygen.temperature.shape
 
 
 @pytest.mark.parametrize(('density', 'coupling', 'shape'),[


wtbarnes:
Here's the current comparison with the IDL code:

image

There are some minor differences, but given the differences in implementation, this isn't too surprising. In particular, it doesn't seem like the IDL implementation transitions smoothly to 1 for the H- and He-like ions. 

What is somewhat surprising is the differences between what is implemented here and the published result (Fig. 2 of Nikolic et al 2018):

image

In particular, the O 4+ (Be-like sequence) appears very different to what is published. I'm not sure why this is or if there is a special case for $N=4$ that I'm not accounting for.

wtbarnes:
The updated comparison (solid is fiasco, dashed is IDL) is below:

image

wtbarnes:
I've now just made this a private method and I'm going to add the docstring info for this feature to the dielectronic recombination rate in a future PR as the user interface for this functionality is going to be via the actual rate calculation.

wtbarnes:
Thanks @pryoung. I did a comparison with your most recent changes and get very good agreement for all of the relevant isoelectronic sequences now,

image

There is a small offset in the Be-like sequence but its below 1% relative difference so I'm not that concerned,

image

PR #423, opened 2026-01-05
Bump minimum Python version to 3.12, add ability to calculate periodic table period

wtbarnes:
This bumps the minimum supported Python version to 3.12 in accordance with SPEC0 and adds testing for Python 3.14.

To support Python 3.14, this PR also gets rid of the dependency on the [`mendeleev`](https://mendeleev.readthedocs.io/en/stable/) package which was previously being used to compute the period of a given element. The reasoning for getting rid of this dependency was twofold: 1. `mendeleev` was not compatible with Python 3.14 and 2. the function used for looking up the period was considerably slower than this simple implementation. In the end, it didn't make sense to pick up this dependency when only a very small part of the package was being used.

This PR makes two API changes: a function in `fiasco.util.periodic_table_period` which computes the period for a given element and a new property `period` on `Ion` that provides the period of the element of that ion.

### DIFF
--- a/fiasco/ions.py
+++ b/fiasco/ions.py
@@ -685,7 +685,9 @@ Using Datasets:
     def _rate_matrix_radiative_decay(self) -> u.Unit('s-1'):
         rate_matrix = u.Quantity(np.zeros((self.n_levels, self.n_levels)), 's-1')
         # Radiative decay into current level from upper levels
-        rate_matrix[self.transitions.lower_level-1, self.transitions.upper_level-1] += self.transitions.A
+        lower_index = self.transitions.lower_level-1
+        upper_index = self.transitions.upper_level-1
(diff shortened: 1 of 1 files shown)


wtbarnes:
It seems as though something has changed from Python 3.13 to 3.14 that causes the contribution function comparison tests to fail for a few of the example ions.

Looking a little more closely at Fe IX 171.073 Å, it seems this is due to differences in the level populations calculation. As an example, here is the level population as a function of temperature for that transition at a constant density of $10^{10}$ $\mathrm{cm}^{-3}$ as calculated using Python 3.13

image

and here is the same calculation done using Python 3.14

image

I do not yet know why these are different or why these would only yield such drastically different results for only a few transntions.

These failures seem to persist across all supported versions of the database for Fe IX 171.073 and 188.496. In v10 and v11, the Ca XV 200.972 test also fails.

PR #426, opened 2026-01-08
Modify ion.n_levels

### LINKED ISSUE #345: Replace all usage of `_wgfa` with `self.transitions`
In `Ion`, when accessing information about the transitions, the `_wgfa` property is often accessed directly. Instead, `.transitions` should be used (which depends on `_wgfa`). The goal here is to minimize the API surface between the CHIANTI-specific low-level API and the higher-level user-facing API.

### LINKED ISSUE #425: Too many levels for N 4 starting at Chianti Database 10
n_4.elvlc was updated between Chianti database version 9 and 10.  n_4.elvlc in database version 10 shows 238 energy levels, but n_4.wgfa only contains transitions between 136.  See this note at the bottom of n_4.* files:

        Note: only the 136 bound levels have been retained.
        
        Produced as part of the 'CHIANTI' atomic data base collaboration by
        Giulio Del Zanna, May 2020

The Ion class reads in all 238 levels, resulting in a rate matrix with many all zero rows, and in turn a singular matrix error when trying to calculate the contribution function for N 4.

As a test, I manually deleted transitions beyond 136  in n_4.elvlc, rebuilt the h5 database, and successfully computed response functions for N 4.

The singular matrix error can be reproduced easily starting with Fiasco 0.6

```
from fiasco import Ion
import astropy.units as u

n_4 = Ion('N 4', 1 * u.MK)
n_4.contribution_function(1e9/(u.cm)**3)
```

jacobdparker:
Fixes #425 
Fixes #345 

The main purpose of this PR is to fix the definition of the number of levels used by the CHIANTI atomic model. Previously, it was assumed that this was just equal to the number of energy levels in the `elvlc` file. However, this logic fails in certain cases (e.g. N IV) where there are not radiative decay rates available for all levels given in the `elvlc` file.

As such, the number of levels is now calculated as the minimum across the highest level available in the `elvlc`, `wgfa`, and `scups` files. This ensures that there is enough information to fully describe the atomic model. This is consistent with what is done in CHIANTI IDL. There is an additional correction if `auto` files are present to ensure that a sufficient number of levels are included to capture the autoionization rates.

As part of this refactor, the number of transitions in `Ion.transitions` is now truncated to include only those transitions between levels in the atomic model. `Ion.levels` still includes all information in the `elvlc` file, i.e. even those levels which are not included in the atomic model.

This also includes a minor fix to the two-photon calculation which takes advantage of the label/configuration properties added in #427.

Other misc. changes:
- Added all `auto` files to test database to ensure the number of levels is always calculated appropriately.
- The two-photon continuum comparison now uses the single-ion model to avoid memory consumption limits on RtD. This has no impact on the two-photon calculation itself.
- Several of the other gallery examples are also adjusted to avoid hitting RtD memory limits
- Ruff configuration moved to separate file (from `pyproject.toml`)

### DIFF
--- a/fiasco/levels.py
+++ b/fiasco/levels.py
@@ -49,7 +49,7 @@ class Levels:
     def __getitem__(self, index):
         # NOTE: This throws an IndexError to stop iteration
         _ = self.level[index]
-        return type(self)(self._elvlc, n_levels=self.level.max(), index=index)
+        return type(self)(self._elvlc, index=index)
 
     def __repr__(self):
         return f"""Level: {self.level}


fiasco/ions.py
```diff
@@ -162,7 +162,23 @@ def n_levels(self):
         Number of energy levels in the CHIANTI model
         """
         try:
-            return len(self.levels)
+            n_levels_elvlc= len(self.levels)
+
+            # some ions don't have .wgfa
+            try:
+                n_levels_wgfa = self._wgfa['upper_level'].max()
```
wtbarnes:
```suggestion
                n_levels_wgfa = self.transitions.upper_level.max()
```

fiasco/ions.py
```diff
@@ -162,7 +162,23 @@ def n_levels(self):
...
+            except KeyError:
+                return 0
+
+            n_levels_scups = self._scups['upper_level'].max()
```
wtbarnes: I think you'll want to first check for the existence of the scups dataset using `self._has_dataset('scups')`. I'm not sure if there are any cases where there is a wgfa file but not a scups file, but it doesn't hurt to guard against that.

jacobdparker:
I took the max between n_levels_scups and n_levels_wgfa assuming that if at least one file refered to an energy level, it should be included in the model.  This only really matters for Ar 18, where if you took the minimum, level 25 would be excluded.  Maybe it should be ...

I find two cases (updated from original post) where n_levels_elvlc is > n_levels_scups and n_levels_wgfa:

N 4

n_levels_elvlc=238

n_levels_wgfa=np.int64(136)

n_levels_scups=np.int64(136)

Ar 6

n_levels_elvlc=202

n_levels_wgfa=np.int64(155)

n_levels_scups=np.int64(155)


I find one case where n_levels_scups > n_levels_wgfa:

Ar 18

n_levels_elvlc=25

n_levels_wgfa=np.int64(24)

n_levels_scups=np.int64(25)

And a few cases where there is now scups or wgfa file:

Ti 22

n_levels_elvlc=64

n_levels_wgfa=0

n_levels_scups=0

Cr 24

n_levels_elvlc=64

n_levels_wgfa=0

n_levels_scups=0

Zn 30

n_levels_elvlc=64

n_levels_wgfa=0

n_levels_scups=0

fiasco/ions.py
```diff
@@ -159,10 +159,28 @@ def _has_dataset(self, dset_name):
     @property
     def n_levels(self):
         """
-        Number of energy levels in the CHIANTI model
+        Number of energy levels in the CHIANTI model.  It is possible this number
+        will not match the number of levels in the elvlc file.
```
wtbarnes:
```suggestion
        will not match the number of levels in the energy level file.
```

fiasco/ions.py
```diff
@@ -159,10 +159,28 @@ def _has_dataset(self, dset_name):
...
+
+            if self._has_dataset('scups'):
+                n_levels_scups = self._scups['upper_level'].max()
+            else:
+                n_levels_scups = 0
+
+            if n_levels_elvlc > n_levels_wgfa or n_levels_elvlc > n_levels_scups:
+                n_levels = np.max([n_levels_scups, n_levels_wgfa])
+            else:
+                n_levels = n_levels_elvlc
```
wtbarnes: Can this be reduced to
```suggestion
            n_levels = min(n_levels_elvlc,  np.max([n_levels_scups, n_levels_wgfa]))
```
or am I misunderstanding the logic here?

wtbarnes:
Ok this has become way more complex than I thought. I think only a few remaining fixes though (which I'm documenting here for myself):

- In some cases, the `_elvlc` property is being used directly which for some ions may have more levels than the model has. Make sure this isn't a problem where it is being used.
- Should we just replace all usage of `elvlc` with `levels` and then truncate `levels` to always have the same number of levels as `n_levels`?
- The two-photon calculation indexes the level populations matrix using `elvlc` which is potentially wrong for some ions. The way the configuration indexing is done is also potentially wrong. Both need fixing
- In the two-photon comparison example, limit to the single-ion model to avoid memory limitations on RTD.

PR #427, opened 2026-01-11
Transition Configurations

### LINKED ISSUE #369: Inconsistent Naming Convention (for level configuration)
When printing the different level configurations for different ions, I realized there is a different naming convention. As an example:

Fe and Ar have the following naming convention (spaces)
```
In [64]: fe[8].levels.configuration[0:5]
Out[64]: 
array(['3s2 3p6', '3s2 3p5 3d', '3s2 3p5 3d', '3s2 3p5 3d', '3s2 3p5 3d'],
      dtype='<U13')
```

S and Ca have the filling naming convention (periods)
```
In [65]: ca[13].levels.configuration[0:5]
Out[65]: 
array(['2s2.2p3', '2s2.2p3', '2s2.2p3', '2s2.2p3', '2s2.2p3'],
      dtype='<U14')
```

Be aware when creating masks using level configurations. I am not sure if this is a Chianti or a fiasco issue. Thank you.

jacobdparker:
Fixes #369 

Expose upper and lower configuration labels as properties
Replace '.' with ' ' in configuration label of certain ions for uniformity
Add transitions.configuration

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,7 @@ dependencies = [
   "astropy>=5.2",
   "fortranformat>=2",
   "h5py>=3.9",
-  "numpy>=1.25,<2.3.0",
+  "numpy>=1.25",
   "plasmapy>=2024.10.0",
 ]
 


fiasco/levels.py
```diff
@@ -192,11 +192,30 @@ def upper_level(self):
...
+    def lower_configuration(self):
+        "Label of the lower level of the transition."
+        idx = vectorize_where(self._levels.level, self.lower_level)
+        configuration = self._levels.configuration[idx]
+        return configuration
+
+    @property
+    def configuration(self):
+        "Each configuration of the transition."
+        return self.lower_configuration + ' - ' + self.upper_configuration
```
wtbarnes:
```suggestion
        return f'{self.upper_configuration} -- {self.lower_configuration}'
```

fiasco/tests/test_ion.py
```diff
@@ -103,6 +103,15 @@ def test_level_energy_parsing(fe10):
...
 
+def test_transition_configuration(ion):
+    transition_index = np.arange(ion.n_transitions)
+    upper_index = ion.transitions.upper_level[transition_index]
+    lower_index = ion.transitions.lower_level[transition_index]
+    upper_config_from_dbase = ion._elvlc['config'][upper_index-1]
+    lower_config_from_dbase = ion._elvlc['config'][lower_index-1]
+    assert np.all(upper_config_from_dbase == ion.transitions.upper_configuration[transition_index])
+    assert np.all(lower_config_from_dbase == ion.transitions.lower_configuration[transition_index])
+    assert np.all(ion.transitions.configuration == (lower_config_from_dbase + ' - ' + upper_config_from_dbase))
```
wtbarnes:
```suggestion
    assert np.all(ion.transitions.configuration == (lower_config_from_dbase + ' -- ' + upper_config_from_dbase))
```

fiasco/levels.py
```diff
@@ -192,11 +192,30 @@ def upper_level(self):
...
 
+    @property
+    def lower_configuration(self):
+        "Label of the lower level of the transition."
+        idx = vectorize_where(self._levels.level, self.lower_level)
+        configuration = self._levels.configuration[idx]
+        return configuration
+
+    @property
+    def configuration(self):
```
wtbarnes: I'm wondering if this should be changed to "label" or something a little more generic. I worry that "configuration" conflicts with the term specific to the filling of shells.

fiasco/levels.py
wtbarnes: I just realized that we are skipping the angular momentum and multiplicity terms here. Preserving the lower/upper configuration properties as-is is fine, but I think this label in particular should be modified to include $^{2s+1}L_J$ for each level as well.

wtbarnes: All of these are existing properties on `Level` so it's just a matter of stitching them together.

PR #441, opened 2026-04-01
Added line_ratio and example with EIS data

### LINKED ISSUE #161: Add example gallery entry for density diagnostics
@dstansby  and I have both been using fiasco to do density diagnostics with EIS data. It would be nice to have a gallery example that shows how to compute a density diagnostic curve and (optionally) map observed ratios to the theoretical intensity ratio curve to estimate the observed densities.

A few useful references for this:
* [Young et al. (2009)](https://ui.adsabs.harvard.edu/abs/2009A%26A...495..587Y/abstract)
* [`eis_chianti_dens_ratio.pro`](https://hesperia.gsfc.nasa.gov/ssw/hinode/eis/idl/chianti/eis_chianti_dens_ratio.pro)

If we wanted to do this with an actual observation, we might also need to install the [`eispac`](https://github.com/USNavalResearchLaboratory/eispac) software.

nabobalis:
This is my attempt with AI to add helpers into fiasco which will allow me to create https://soho.nascom.nasa.gov/solarsoft/iris/idl/sao/util/tian/iris_ne.pro inside irispy.

Maybe fixes https://github.com/wtbarnes/fiasco/issues/161


examples/user_guide/plot_jet_density_figure.py
```diff
@@ -0,0 +1,131 @@
...
+
+import fiasco
+
+quantity_support()
+
+###############################################################################
+# First download the precomputed EIS products. These two FITS files are hosted
+# in a separate data repository so that this example does not need to run the
+# EIS fitting step with `eispac <https://eispac.readthedocs.io/en/latest/>`__.
+data_base_url = 'https://media.githubusercontent.com/media/nabobalis/data/main/fiasco'
```
nabobalis: Update URL when PR merged into sunpy/data

fiasco/tests/data/eis_fe_xiii.readme
wtbarnes, line 1: What is this supposed to be?

fiasco/tests/data/eis_fe_xiii.sav
wtbarnes, line 1: This should be be removed. See https://github.com/wtbarnes/fiasco/blob/3194c1a5b894a79890927e340b8567f9843f1eeb/fiasco/tests/idl/test_idl_ioneq.py#L26 for an example of how to add an IDL comparison test.

fiasco/tests/test_fiasco.py
```diff
@@ -31,8 +30,38 @@ def test_get_isoelectronic_sequence(hdf5_dbase_root):
...
 def test_proton_electron_ratio(hdf5_dbase_root):
     t = np.logspace(4, 9, 100) * u.K
-    # NOTE: this number will not be accurate as we are using only a subset of
-    # the database
     pe_ratio = fiasco.proton_electron_ratio(t, hdf5_dbase_root=hdf5_dbase_root)
-    assert type(pe_ratio) is u.Quantity
-    assert pe_ratio.shape == t.shape
+    assert np.all(np.isfinite(pe_ratio))
+    assert np.all(pe_ratio > 0)
+    assert np.ptp(pe_ratio.value) > 0
```
wtbarnes: What is the point of this test?

fiasco/tests/test_fiasco.py
```diff
@@ -31,8 +30,38 @@ def test_get_isoelectronic_sequence(hdf5_dbase_root):
 
 def test_proton_electron_ratio(hdf5_dbase_root):
     t = np.logspace(4, 9, 100) * u.K
-    # NOTE: this number will not be accurate as we are using only a subset of
-    # the database
     pe_ratio = fiasco.proton_electron_ratio(t, hdf5_dbase_root=hdf5_dbase_root)
-    assert type(pe_ratio) is u.Quantity
-    assert pe_ratio.shape == t.shape
+    assert np.all(np.isfinite(pe_ratio))
```
wtbarnes:
```suggestion
    assert np.isfinite(pe_ratio).all()
```

wtbarnes (commented):
I've left a few detailed comments, but am holding off on a full line-by-line review as a I have a couple of high-level comments.

The functionality for interpolating between the theoretical and observed ratios is out of scope for fiasco. It is easy enough to do this in 2-3 lines in the example gallery without a dedicated function in fiasco. Additionally, the theoretical line ratio method should be a separate function (it can go in `fiasco/fiasco.py`). I am worried about overcrowding of the `Ion` namespace and this is more of an "analysis" function rather than something directly related to properties of the ion.

fiasco/ions.py
```diff
@@ -236,6 +236,28 @@ def transitions(self):
         "A `~fiasco.Transitions` object holding the information about transitions for this ion."
         return Transitions(self.levels, self._wgfa, n_levels=self.n_levels)
 
+    def _transition_indices(self, transitions):
```
wtbarnes: Get rid of this

fiasco/ions.py
```diff
@@ -1154,6 +1176,52 @@ def contribution_function(self, density: u.cm**(-3), **kwargs) -> u.cm**3 * u.er
...
+        ----------
+        numerator, denominator : `~astropy.units.Quantity`
+            Transition wavelength(s) for the numerator and denominator. If multiple
+            transitions are provided, their contribution functions are summed before
+            taking the ratio.
+        density : `~astropy.units.Quantity`
+            Electron number density.
+        temperature : `~astropy.units.Quantity` or `None`, optional
+            Temperature grid on which to evaluate the ratio. If `None`, the
+            temperature grid of the current ion instance is used.
```
wtbarnes: Get rid of this input. The result should always be evaluated on the `ion` temperature grid.

fiasco/ions.py
```diff
@@ -1154,6 +1176,52 @@ def contribution_function(self, density: u.cm**(-3), **kwargs) -> u.cm**3 * u.er
...
+                   **kwargs):
+        r"""
+        Theoretical line ratio for one or more bound-bound transitions as a function of density.
+
+        Parameters
+        ----------
+        numerator, denominator : `~astropy.units.Quantity`
+            Transition wavelength(s) for the numerator and denominator. If multiple
+            transitions are provided, their contribution functions are summed before
+            taking the ratio.
```
wtbarnes: It should be possible to provide either an array of wavelengths or an array of strings representing the transition labels (see https://github.com/wtbarnes/fiasco/blob/3194c1a5b894a79890927e340b8567f9843f1eeb/fiasco/levels.py#L263). The latter is a more exact way to specify a transition. In the case of passing a label, the appropriate index can be found through a string equality check.

fiasco/ions.py
```diff
@@ -1154,6 +1176,52 @@ def contribution_function(self, density: u.cm**(-3), **kwargs) -> u.cm**3 * u.er
...
+            Passed to :meth:`~fiasco.Ion.contribution_function`.
+
+        Returns
+        -------
+        `~astropy.units.Quantity`
+            The line ratio with shape ``(n_temperature, n_density)`` for independent
+            density arrays or ``(n_temperature, 1)`` for ``couple_density_to_temperature=True``.
+            Points where the denominator vanishes are returned as `~numpy.nan`.
+        """
+        ion = self if temperature is None else self._new_instance(temperature=temperature)
```
wtbarnes:
```suggestion
```

fiasco/tests/data/eis_fe_xiii.readme
nabobalis: Some text to explain how and when the sav file was created

fiasco/tests/test_fiasco.py
nabobalis: Its better than the previous version?

fiasco/tests/data/eis_fe_xiii.sav
nabobalis: You would rather have people use IDL to run the tests over test files?

fiasco/fiasco.py
```diff
@@ -165,3 +166,76 @@ def proton_electron_ratio(temperature: u.K, **kwargs):
...
+
+
+@u.quantity_input(density=u.cm**(-3))
+def line_ratio(ion,
+               numerator,
+               denominator,
+               density: u.cm**(-3),
+               **kwargs):
+    """
+    Theoretical line ratio for one or more bound-bound transitions as a function of density.
```
jwreep: Line ratios can also be used for measurements of temperature, abundance, and magnetic field strength.

Perhaps change the function name to "line_ratio_density" or something to make it clear which one this function uses.  (Or generalize the function to include the other cases?)

nabobalis: I can rename it to line_ratio_density for now. If Will wants, I can take a stab at the other use-cases for it. But now, I am just focused on getting this merged so I can use it for IRIS.

fiasco/tests/data/eis_fe_xiii.sav
wtbarnes: You don't have to. The results are cached and reused if present. The IDL code is only run if the cached data are not available and an IDL installation is available. From what I can tell, your vibeocoded solution may have removed part of this logic which is why you may think that you have to run IDL.

fiasco/fiasco.py
wtbarnes: As is, it actually already can be used as a temperature line ratio function so I've just renamed it to `line_ratio` and made it clear in the docstring it is a function of temperature and density.

examples/user_guide/plot_jet_density_figure.py
wtbarnes: done

examples/user_guide/plot_jet_density_figure.py
```diff
@@ -0,0 +1,101 @@
...
+from astropy.visualization import quantity_support
+
+import fiasco
+
+quantity_support()
+
+###############################################################################
+# First download and read in the precomputed EIS products. These two FITS files are hosted
+# in a separate data repository so that this example does not need to run the
+# EIS fitting step with [eispac](https://eispac.readthedocs.io/en/latest/)_.
```
nabobalis:
```suggestion
# EIS fitting step with `eispac <https://eispac.readthedocs.io/en/latest/>`__.
```
Markdown by mistake

nabobalis:
LETS GO

Issue #274, opened 2024-03-18
Enforce a consistent interpolation method


wtbarnes:
Data often need to be interpolated to a particular grid, usually in temperature space. Initially, I used [`interp1d`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp1d.html) for everything, but that function is now considered "legacy" and thus should be avoided.

Additionally, for reasons explained more fully in #223, I've used a PCHIP interpolation method in log-log space for particular pieces of data.

Now, in #260, `CubicSpline` is being used. 

We need to come up with a policy for which interpolation method we use by default and stick to a single method unless there is a compelling reason to deviate from that method. Once a decision is made, this should probably be put in the docs somewhere.

jwreep:
I used `CubicSpline` in #260 simply because it closely matches the IDL code.  Happy to switch if another is preferred.

wtbarnes:
Actually, I think your approach is the right one. We should replace `interp1d` with `CubicSpline` everywhere and use `PCHIPInterpolator` where necessary.

wtbarnes:
I changed the title a bit to reflect that I don't think there is a one-size fits all solution to this problem. We should instead provide some brief guidance on which method to choose. This could be a topic guide in the docs.

PR #444, opened 2026-05-05
Make ionization and recombination components functions, optionally return level-resolved rates

### LINKED ISSUE #384: Update ionization rate calculation to include level-resolved rates
In v11, the overall ionization rate can be computed as the sum of the level-resolved ionization rates weighted by the populations of those levels. The levels for which these ionization rates are provided are the so-called metastable levels. See section 2.1 and 2.4 of [Dufresne et al. (2024)](https://ui.adsabs.harvard.edu/abs/2024ApJ...974...71D/abstract) for more details.

One thing that will need to be decided is how to incorporate the density as these updated rates are now density-dependent since they depend on the level populations calculation. However, the ionization rate is currently implemented as a property such that it does not take any inputs. There are two options:

1. Turn `ionization_rate` into a method that takes a density
2. Make `density` an (optional) input to the `Ion` class

Even if using option 2, `ionization_rate` could still be a method that takes an argument saying whether to calculate the ionization rate using the level-resolved rates or not.

Before this can be completed, #374, #376, #377 need to be completed.

### LINKED ISSUE #385: Update recombination rate calculation to include level-resolved rates
In v11, the recombination rate can optionally be calculated as a sum of the level-resolved rates weighted by the level populations of these levels. This is only applicable for those ions which have so-called "advanced models". There are level-resolved rates for both radiative and dielectronic recombination. In the case of dielectronic recombination, the density-dependent suppression factors discussed #233 are also incorporated.

The API changes discussed in #384 to account for density as an input also apply here and the same approach (whatever that is) should be used.

Before this can be completed, #374, #233, #379, #378 need to be completed.

### LINKED ISSUE #415: Modify ionization and recombination rates to be optionally density dependent
Currently, the ionization and recombination rates are properties on `Ion` because they are zero-density and only temperature-dependent. However, with #384 and #385 and #411, these rates may also be density-dependent. As such, these properties need to be turned into methods that optionally take a density.

### LINKED ISSUE #416: Account for density-dependent suppression in dielectronic recombination
After #411 is merged, the dielectronic recombination calculation on `Ion` needs to be modified to optionally account for density-dependent suppression. This can be done alongside #415. 

As part of this, the docstring should also be modified to document what this suppression does, including details of how it is implemented with appropriate references.

wtbarnes:
Fixes #384 
Fixes #385 
Fixes #415
Fixes #416

## Remaining TODOs

- [x] Return level-resolved ionization rates
- [x] Return level-resolved recombination rates
- [x] Add function for selecting a model atmosphere
- [x] Add methods for calculating CT ionization and recombination rates
- [ ] Implement metastable level estimation (see [`metastable_levels.pro`](https://github.com/chianti-atomic/chianti-idl/blob/main/utilities/metastable_levels.pro) in the IDL code) for neutrals
- [ ] Implement Burgess-Chidichimo ionization rates in cases where there are no level-resolved ionization rate data available. See [`ch_burgchid_rate.pro`](https://github.com/chianti-atomic/chianti-idl/blob/d5f2bd3826418c6d4c0f4f1b6ac4211a798b5f89/ionrec/ch_burgchid_rate.pro). This should probably be a private method on `Ion` as it has limited public utility and should not be used in place of other ionization rates.
- [ ] Modify level populations method to accommodate a limited number of levels.
- [ ] Implement full density-dependent rate calculation, including all rate components. 
  - Not clear yet where this should go
  - Maybe for now, just put the logic in the respective ionization and recombination rate methods.
- [ ] Allow `Element.ionization_equilibrium` to accept a density and atmosphere model.


wtbarnes:
Some working notes for myself

- Should have separate methods for charge transfer ionization/recombination
- These can be added into total for ionization/recombination if model atmosphere is considered
- Should ionization/recombination rate methods be able to return the advanced model rates?
- These have the downside of needing to repeatedly calculate the level populations...
- In the case of the recombination rates, there is an added complication because using the `density=` kwarg might mean
  - You want the zero-density rates multiplied by the density-dependent suppression factor or
  - You want the effective density-dependent rate calculated from the level-resolved rates multiplied by the suppression factor
  - You want the effective density-dependent rate calculated from the level-resolved rates without the suppression factor
  - Need to decide the logic on how best to allow for these different scenarios.
- If not, then are they just all calculated in `Element.ionization_equilibrium`?

wtbarnes:
There are properties related to the older treatment of the level-resolved ionization and recombination rate data (cilvl and reclvl files, respectively) that I've called `_level_resolved_ionization_rate` and `_level_resolved_recombination_rate` that have an unfortunate namespace clash with the functionality I'm adding here. 

https://github.com/wtbarnes/fiasco/blob/43ad8655381355cca6364f6507d2eab81cbdb136/fiasco/ions.py#L997-L1019

These should just be folded into the `_population_correction` method as that is the only place they are used anyway. 

https://github.com/wtbarnes/fiasco/blob/43ad8655381355cca6364f6507d2eab81cbdb136/fiasco/ions.py#L1022

These are not meant to be used on their own and should not be confused with the new advanced model data.

wtbarnes:
In trying to match the level-resolved ionization rates in IDL, I'm finding some peculiarities in the IDL interpolation. In the IDL code, this is done using the `interpol` function with `/spline`. The relevant lines for interpolating the level-resolved rates are here: https://github.com/chianti-atomic/chianti-idl/blob/d5f2bd3826418c6d4c0f4f1b6ac4211a798b5f89/ionrec/ch_ioniz_rate_lr.pro#L241-L250. Internally, this seems to use `spl_interp` which in other instances has been matched well with `scipy.interpolate.CubicSpline` (see https://github.com/wtbarnes/fiasco/pull/297#issuecomment-2339028303). 

The comparison for the direct-ionization rates matches to within 1% except near the lower boundary where the single-precision in IDL leads to the rates at the lower data boundary being truncated at $10^{-50}$. The Python implementation matches the actual data much more closely, but the differences are minimal.

image

The excitation auto-ionization rates on the other hand show much larger differences, up to 25% over the relevant temperature range. They also show the same pathology at the lower end of the temperature range.

image

I suspect none of this will be that important as the direct ionization rates will be larger most of the time and the recombination rates will also be dominant at the temperatures where this matters, but it still annoying. 

See also #274

wtbarnes:
There are issues with the interpolation of the IDL CT ionization and recombination rates that may cause issues in the comparison. See chianti-atomic/chianti-idl#21. I believe that I've resolved this on the Python side by setting the fill values of the interpolation at the extrema appropriately.

For example, the following shows the interpolated ionization and recombination CT data for Si III, comparing the IDL and fiasco cases:

image

image

PR #446, opened 2026-05-22
Update zenodo.json and contributing guide

### LINKED ISSUE #445: Update `.zendo.json` to include new contributors
There are several new contributors that have not been included in the `.zenodo.json` file. This should be fixed so that the Zenodo record accurately reflects the contributors to the package.

jwreep:
Fixes #445 

See my comment, though, on that issue.  Perhaps we should add a simple rule that first-time contributors should add their info to the file.  I could update the docs to include a sentence to that effect if you agree with it.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -38,12 +38,12 @@
     {
       "name": "Laura Hayes",
       "orcid": "0000-0002-6835-2390",
-      "affiliation": "European Space Research and Technology Centre, European Space Agency"
+      "affiliation": "Dublin Institute for Advanced Studies"
     },
     {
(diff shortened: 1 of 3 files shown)


wtbarnes:
>  I could update the docs to include a sentence to that effect if you agree with it.

Yes, that's a great idea. If you could add that it'd be great!

jwreep:
I've edited the contributing guide to mention this briefly.  Looks like the docs build is failing, but I see you've got a PR open to address that.

wtbarnes:
Thanks! Yes the CI is a bit broken on all fronts but I'll fix it in the other PR and then rebase this one.

PR #450, opened 2026-07-11
Fix level_populations matrix solve shapes for numpy<2 compatibility

### LINKED ISSUE #447: Bug in contribution function?
I was running the tutorial on calculating contribution functions, and hit a snag.  Not sure I understand the root of the issue, but this is with **fiasco version 0.8.1 and dbase version 8.0.7**.  It looks like the issue arises during the level population calculation, so possibly a change relating to the updates there with newer versions of the database is causing a problem?

I ran the following lines (copied verbatim from the how-to):

```
import astropy.units as u
import matplotlib.pyplot as plt
import numpy as np

from astropy.visualization import quantity_support

from fiasco import Ion

quantity_support()

Te = np.geomspace(0.1, 10, 26) * u.MK
ne = 1e8 * u.cm**-3

ion = Ion('O 5+', Te)
print(ion)

contribution_func = ion.contribution_function(ne)
```

which gives the following traceback:
```
WARNING: No autoionization or level-resolved radiative recombination data available for O 6. Using single-ion model for level populations calculation. To silence this warning, set use_two_ion_model=False. [fiasco.ions]
WARNING: No proton data available for O 6. Not including proton excitation and de-excitation in level populations calculation. [fiasco.ions]

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[4], line 1
----> 1 contribution_func = ion.contribution_function(ne)

File ~/anaconda3/lib/python3.12/site-packages/fiasco/util/decorators.py:32, in needs_dataset.<locals>.decorator.<locals>.func_wrapper(*args, **kwargs)
     27     except KeyError:
     28         raise MissingDatasetException(
     29             f"{n} dataset missing for {getattr(obj, 'ion_name', obj)}."
     30         )
---> 32 return func(*args, **kwargs)

File ~/anaconda3/lib/python3.12/site-packages/astropy/units/decorators.py:316, in QuantityInput.__call__.<locals>.wrapper(*func_args, **func_kwargs)
    314 # Call the original function with any equivalencies in force.
    315 with equiv_context:
--> 316     return_ = wrapped_function(*func_args, **func_kwargs)
    318 # Return
    319 ra = wrapped_signature.return_annotation

File ~/anaconda3/lib/python3.12/site-packages/fiasco/ions.py:1135, in Ion.contribution_function(self, density, **kwargs)
   1083 r"""
   1084 Contribution function :math:`G(n_e,T)` for all transitions.
   1085 
   (...)   1132 level_populations
   1133 """
   1134 couple_density_to_temperature = kwargs.get('couple_density_to_temperature', False)
-> 1135 populations = self.level_populations(density, **kwargs)
   1136 if couple_density_to_temperature:
   1137     term = self.ionization_fraction / density

File ~/anaconda3/lib/python3.12/site-packages/astropy/units/decorators.py:316, in QuantityInput.__call__.<locals>.wrapper(*func_args, **func_kwargs)
    314 # Call the original function with any equivalencies in force.
    315 with equiv_context:
--> 316     return_ = wrapped_function(*func_args, **func_kwargs)
    318 # Return
    319 ra = wrapped_signature.return_annotation

File ~/anaconda3/lib/python3.12/site-packages/fiasco/ions.py:639, in Ion.level_populations(self, density, include_protons, include_level_resolved_rate_correction, couple_density_to_temperature, use_two_ion_model)
    637 b = np.zeros(c_matrix.shape[2:])
    638 b[-1] = 1.0
--> 639 pop = np.linalg.solve(c_matrix.value, b)
    640 pop[pop<0] = 0.0
    641 np.divide(pop, pop.sum(axis=1)[:, np.newaxis], out=pop)

File ~/anaconda3/lib/python3.12/site-packages/numpy/linalg/linalg.py:409, in solve(a, b)
    407 signature = 'DD->D' if isComplexType(t) else 'dd->d'
    408 extobj = get_linalg_error_extobj(_raise_linalgerror_singular)
--> 409 r = gufunc(a, b, signature=signature, extobj=extobj)
    411 return wrap(r.astype(result_t, copy=False))

ValueError: solve: Input operand 1 does not have enough dimensions (has 1, gufunc core with signature (m,m),(m,n)->(m,n) requires 2)
```

Leonard013:
Fixes #447.

**Symptom.** `Ion.contribution_function()` → `Ion.level_populations()` raises `ValueError: solve: Input operand 1 does not have enough dimensions (has 1, gufunc core with signature (m,m),(m,n)->(m,n) requires 2)` under `numpy < 2` (e.g. 1.26.4), while succeeding under `numpy >= 2`. This matches @jwreep's finding in the thread: identical fiasco and CHIANTI versions, differing only in numpy.

**Root cause.** `level_populations` builds a stack of coefficient matrices `c_matrix.value` of shape `(n_temperature, n_levels, n_levels)` and solves it against a right-hand side that pins the normalization row. That right-hand side was `b = np.zeros(c_matrix.shape[2:])` — a 1D array of shape `(n_levels,)`. `numpy.linalg.solve`'s gufunc dispatch treats a 1D `b` as a broadcastable stack-of-vectors only when `b.ndim == a.ndim - 1`; here `1 != 2`, so it falls to the `(m,m),(m,n)->(m,n)` matrix signature. numpy>=2.0 added a special case that broadcasts a genuinely-1D `b` across the batch (so it works there), but numpy<2 has no such case and raises. Since `pyproject.toml` declares `numpy>=1.25`, this crashes for any supported user on numpy 1.25/1.26.

**Fix.** Give `b` an explicit trailing singleton axis, shape `(n_temperature, n_levels, 1)`, so `solve` dispatches on the same signature under both numpy majors, then squeeze the trailing axis off the result. This is numerically identical to the previous behaviour on numpy>=2 and removes all version-dependent dispatch.

Note: the naive alternative of broadcasting `b` to `(n_temperature, n_levels)` is *not* correct — under numpy>=2 that shape is reinterpreted as the `(m,n)` core block and either errors (when `n_temperature != n_levels`) or returns a wrong-shaped result (when they coincide). And pinning a `numpy>=2` floor alone would not fix the bug for the currently-declared `numpy>=1.25` support range — it would be a breaking dependency bump rather than a fix. The shape correction is the necessary change; a floor bump would only be optional belt-and-suspenders.

**Verification** (macOS arm64, two virtualenvs):

- `fiasco/tests/test_ion.py::test_level_populations` and `::test_contribution_function` already exercise the vulnerable shape (100-point temperature grid vs real `n_levels`) and assert hardcoded numeric populations, so they are genuine numerical-equivalence checks. On `numpy==1.26.4` they fail before this change with the reported `ValueError` and pass after; on `numpy==2.5.1` they pass before and after. (`pytest --pyargs fiasco --ascii-dbase-root ~/.chianti -k "level_populations or contribution_function"`)
- A standalone, CHIANTI-independent check confirms the fixed `(n_temperature, n_levels, 1)` right-hand side matches an independent per-temperature `np.linalg.solve` loop exactly, under both numpy 1.26.4 and 2.5.1.
- `ruff check fiasco/ions.py` is clean.

Optional follow-up (out of scope here): fiasco's declared `numpy>=1.25` floor is never exercised in CI (no `oldestdeps` tox env), which is why this shipped silently; an oldest-deps job would guard against this class of regression.

### DIFF
--- a/changelog/450.bugfix.rst
+++ b/changelog/450.bugfix.rst
@@ -1,1 +1,0 @@
-Fix a ``ValueError`` in `fiasco.Ion.level_populations` (and therefore `fiasco.Ion.contribution_function`) when running with ``numpy < 2``, caused by the shape of the right-hand side passed to `numpy.linalg.solve`.
--- a/changelog/450.bugfix.rst
+++ b/changelog/450.bugfix.rst
@@ -1,0 +1,1 @@
+Fix a ``ValueError`` in `fiasco.Ion.level_populations` (and therefore `fiasco.Ion.contribution_function`) when running with ``numpy < 2``, caused by the shape of the right-hand side passed to `numpy.linalg.solve`.
--- a/fiasco/ions.py
+++ b/fiasco/ions.py
@@ -634,9 +634,15 @@ Using Datasets:
                 c_matrix = self._build_coefficient_matrix(d, include_protons=include_protons)
             # Invert matrix
             c_matrix[:, -1, :] = 1.*c_matrix.unit
-            b = np.zeros(c_matrix.shape[2:])
-            b[-1] = 1.0
-            pop = np.linalg.solve(c_matrix.value, b)
+            # NOTE: c_matrix is a stack of matrices, one per temperature, with shape
+            # (n_temperature, n_levels, n_levels). Give the right-hand side an explicit
+            # trailing axis, shape (n_temperature, n_levels, 1), so that np.linalg.solve
+            # dispatches on the same (m,m),(m,n)->(m,n) signature under both numpy<2 and
+            # numpy>=2. A 1D right-hand side is only broadcast over the stack by numpy>=2,
+            # and raises on numpy<2 (see gh#447).
+            b = np.zeros(c_matrix.shape[:-1] + (1,))
+            b[..., -1, 0] = 1.0
+            pop = np.linalg.solve(c_matrix.value, b)[..., 0]
             pop[pop<0] = 0.0
             np.divide(pop, pop.sum(axis=1)[:, np.newaxis], out=pop)
             # Apply ionization/recombination correction


wtbarnes:
Thanks for this fix! In the interest of transparency, I've removed Claude as a co-author on your commit as I do not want Claude showing up as a contributor on this repository. Use of AI for PRs on this repository is acceptable.

I've removed the changelog as well as we don't currently don't make use of them when it comes time for a release. Apologies that this wasn't clear from the contributing guidelines. Thanks again for this fix and apologies for taking so long to get around to addressing it.

PR #466, opened 2026-08-10
Fast X: level populations edition

nabobalis:
Test to see if Claude MD can speed up some code in this repo.
According to Claude (unverified), these are the speeds up.

```
    ┌───────────────────────────┬────────────────┬────────┐
    │ Case                      │ Original       │ Now    │
    ├───────────────────────────┼────────────────┼────────┤
    │ Fe XXIII two-ion (21T×9n) │ 73.5 s         │ 5.2 s  │
    ├───────────────────────────┼────────────────┼────────┤
    │ Ca XVIII two-ion          │ 74.3 s         │ 6.4 s  │
    ├───────────────────────────┼────────────────┼────────┤
    │ O VI default              │ 73.3 s         │ 6.9 s  │
    ├───────────────────────────┼────────────────┼────────┤
    │ Fe XII single-ion         │ 9.9 s          │ 3.4 s  │
    ├───────────────────────────┼────────────────┼────────┤
    │ Coupled n–T               │ 14.5 s         │ 1.4 s  │
    ├───────────────────────────┼────────────────┼────────┤
    │ proton_electron_ratio     │ ~39 s* / 1.4 s │ 0.38 s │
    └───────────────────────────┴────────────────┴────────┘
```

Now whether the code is good, is another question.

PR #471, opened 2026-09-18
Use the cached proton-electron ratio in `Ion.emissivity` and make it settable

### LINKED ISSUE #470: `emissivity` recomputes `proton_electron_ratio` on every call instead of using the cached property; no way to pass a precomputed ratio
### Summary

`proton_electron_ratio` is expensive (about 12 s per call on my machine, see below), and it is recomputed from scratch far more often than it needs to be:

1. `Ion.emissivity` calls the module-level `proton_electron_ratio(self.temperature, **self._instance_kwargs)` directly rather than using the `Ion.proton_electron_ratio` cached property that already exists on the instance. Since `Ion.intensity` and `IonCollection.spectrum` go through `emissivity`, a collection of N ions walks the entire database N times, and calling `emissivity` twice on the same ion walks it twice.
2. There is no way to hand an ion a precomputed ratio. `abundance` and `ionization_fraction` have setters that accept a value instead of a dataset name; `proton_electron_ratio` does not, so the only workaround is writing into `ion.__dict__` to prime the cached property.

I think both are omissions rather than design decisions: the cached property clearly exists for this purpose, and the ratio depends only on the temperature array and the abundance/ioneq datasets, which the instance already knows.

### Where the time goes

On fiasco 0.8.2 with the full CHIANTI database, 495 ions, 61 temperatures:

| | time |
|---|---|
| `proton_electron_ratio(T)` | 11.6 s |
| second identical call | 11.2 s |
| `Element("Fe", T)` (27 ions) | 0.43 s |

Profiling `Element("Fe", T)` shows roughly three quarters of each `Ion.__init__` is plasmapy particle construction via `atomic_symbol` / `atomic_number` (called ~11 times per ion, uncached), and most of the rest is repeated HDF5 opens in `_has_dataset`, `_ion_fraction`, `_ip`, plus `check_database` and `list_ions` on every instantiation. The physics only needs the abundance table and the ioneq table, and every ion in the database shares the same 101-point ioneq temperature grid, so the whole ratio is a sum over two small arrays.

I see that #466 already addresses most of this (memoized plasmapy lookups and `list_ions`, and a rewrite of `proton_electron_ratio` that instantiates one ion per element and reads the ioneq data directly). That should cut the cost of a single call by a large factor. But as far as I can tell from the diff it does not change `emissivity`, so the ratio would still be recomputed once per ion in `IonCollection.spectrum`, and there is still no setter.

### Proposal

Two small changes, which might fit naturally into #466 or a follow-up:

1. In `Ion.emissivity`, replace the direct call with `self.proton_electron_ratio` so the instance cache is used.
2. Add a `proton_electron_ratio` setter, mirroring the `abundance` setter, so a caller computing spectra for many ions over one temperature grid can compute the ratio once with `fiasco.proton_electron_ratio` and assign it to each ion. `IonCollection.spectrum` could then compute it once and pass it to every ion in the collection.

Happy to open a PR for either if that would help.

### Reproduction

fiasco 0.8.2, full CHIANTI database, O V so that the level population solve is cheap enough for the ratio to stand out:

```python
import time
import astropy.units as u
import numpy as np
import fiasco

T = np.logspace(4, 7, 61) * u.K
ion = fiasco.Ion("O 5", T)
n = 1e9 / u.cm**3

def timed(label, f):
    t0 = time.perf_counter(); f(); print(f"{label:<40s}{time.perf_counter() - t0:6.1f} s")

timed("ion.proton_electron_ratio (first)", lambda: ion.proton_electron_ratio)
timed("ion.proton_electron_ratio (second)", lambda: ion.proton_electron_ratio)
timed("contribution_function", lambda: ion.contribution_function(n))
timed("emissivity (first)", lambda: ion.emissivity(n))
timed("emissivity (second)", lambda: ion.emissivity(n))
```

```
ion.proton_electron_ratio (first)         12.0 s
ion.proton_electron_ratio (second)         0.0 s
contribution_function                      3.9 s
emissivity (first)                        15.2 s
emissivity (second)                       15.0 s
```

The cached property works: `contribution_function` picks it up through `level_populations` and costs 4 s. `emissivity` bypasses it and costs the full 12 s again on every call.

jacobdparker:
Fixes #470

`Ion.emissivity` called the module-level `proton_electron_ratio` directly on every call, walking the entire database each time, instead of using the cached property on the instance. Since `intensity` and `IonCollection.spectrum` go through `emissivity`, a collection of N ions paid for the walk N times.

Changes:

- `Ion.emissivity` now uses `self.proton_electron_ratio`.
- `Ion.proton_electron_ratio` is now a lazily computed property with a setter, in the same style as `abundance` and `ionization_fraction`, so a ratio computed once with `fiasco.proton_electron_ratio` can be assigned to every ion sharing a temperature array. Setting it does not change `_instance_kwargs`, so ions created through `next_ion` and friends compute their own as before.
- Two tests: the setter accepts a scalar or an array of the temperature's shape, and `emissivity` reflects a ratio that was set.

Timing with the full database, O V, 61 temperatures, before and after:

| | before | after |
|---|---|---|
| `emissivity` (first call) | 15.2 s | 15.9 s |
| `emissivity` (second call) | 15.0 s | 3.1 s |

The first call still computes the ratio once; #466 makes that part cheaper. Results are unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


wtbarnes:
I started by just refactoring the setter logic for the proton-electron ratio but that led into a whole other set of problems as well as me discovering some pretty needless slowdowns in both the ratio function as well as how some of the properties on `IonBase` are built.

In summary:
- you can now pass in a value for `proton_electron_ratio` to `Ion`
- you can also set the value of `proton_electron_ratio` on an already constructed `Ion` object
- the `proton_electron_ratio` function is about 10x faster now.
- constructing an `Ion` object may be *slightly* slower now by default, but you can easily get around this by passing in a precomputed `proton_electron_ratio`

